PR #3, opened 2017-11-30
Add unit tests for utility functions

mimischi:
The PR adds the first unit tests for the utility functions found in `utils.py`.

## Todo

- ~~Add test for `utils.guess_ncores`~~

PR Checklist
------------
 - [x] CHANGELOG updated?


setup.cfg
```diff
@@ -1,3 +1,13 @@
...
-ignore = N806, N803, N802, I100, I101, I201
\ No newline at end of file
+ignore = N806, N803, N802, I100, I101, I201
+
+[aliases]
+test=pytest
+
+[tool:pytest]
+testpaths = mdbenchmark/tests
+addopts = --cov=./
```
kain88-de: please remove coverage line. we don't want to do this everytime.

setup.cfg
```diff
@@ -1,3 +1,13 @@
...
 max-line-length = 100
-ignore = N806, N803, N802, I100, I101, I201
\ No newline at end of file
+ignore = N806, N803, N802, I100, I101, I201
+
+[aliases]
+test=pytest
+
+[tool:pytest]
+testpaths = mdbenchmark/tests
```
kain88-de: pytest can figure this out itself. I would remove this line.

kain88-de: that way it's also more future resilient to changes in the file structure!

setup.cfg
```diff
@@ -1,3 +1,13 @@
...
+addopts = --cov=./
+
+[coverage:run]
+omit = mdbenchmark/tests/*
```
kain88-de: we want to know if our tests are all run!

setup.py
```diff
@@ -28,8 +28,16 @@
...
+        'mdsynthesis',
+        'click',
+        'jinja2',
+        'pandas',
+        'matplotlib',
+        'python-Levenshtein',
+        'xdg<2',
     ],
     entry_points={'console_scripts': ['mdbenchmark=mdbenchmark.cli:cli']},
+    setup_requires=['pytest-runner'],
```
kain88-de: setup doesn't need pytest

setup.py
```diff
@@ -28,8 +28,16 @@
...
+    tests_require=['pytest', 'mock', 'pytest-cov'],
```
kain88-de: please remove coverage for the tests. This is only useful for CI runs and occasionally on a dev machine. Adding coverage usually makes tests slower to run.

mdbenchmark/tests/test_utils.py
```diff
@@ -0,0 +1,96 @@
...
+def test_calc_slope_intercept():
+    """Test `calc_slope_intercept()`"""
+    x1, y1 = [1, 1]
+    x2, y2 = [2, 2]
+    slope = (y2 - y1) / (x2 - x1)
+    intercept = y1 - (x1 * slope)
+
+    slope_intercept = calc_slope_intercept(x1, y1, x2, y2)
+
+    np.testing.assert_array_almost_equal(slope_intercept,
```
kain88-de: use top level import of assert function. Also `assert_almost_equal` is more strict then the `assert_array_*` functions. Don't ask me why numpy devs did this.

setup.py
kain88-de: is the mock really needed?

.travis.yml
```diff
@@ -15,6 +15,5 @@ install:
   - pip install --upgrade numpy cython
   - pip install -e .
 
-
 script:
-  - true
+  - python setup.py test
```
kain88-de: if you add tests with coverage information please enable coveralls or codecov as well

mimischi: So remove the coverage completely or keep it? Also you'd need to enable one of the integrations for the `bio-phys` organization, I can't.

setup.py
mimischi: We need `mock` to test the responses from system packages like `socket.gethostname`. This way we can test it on all machines. If we wouldn't do that, then we'd need to write tests specifically for Travis-CI and these would fail on any other machines.
See for example: https://github.com/bio-phys/MDBenchmark/pull/3/files#diff-4c190294c14d18b75f1b5a206c73944cR32

mimischi: Oh wait. I just realized that we're anyway using the `monkeypatch` fixture from `pytest`. It's possible that we can actually get rid of the `mock` dependency then.

.travis.yml
kain88-de: sure? Codecov doesn't seem to need any special permissions. You have to change the travis config. Have a look at `MDAnalysis/pytng`

mimischi: Just checked, yes. It's exactly like with Travis-CI. You grant the access for the whole organization once and then all repositories can use it afterwards.

setup.cfg
mimischi: Isn't that the job of `pytest`? It will give you a list of tests that ran and those that were skipped/failed..?

setup.py
mimischi: I added that in case we use `python setup.py test` to run the tests, as I had some problems with setting up `pytest` initially.

.travis.yml
kain88-de: should work now

.travis.yml
```diff
@@ -14,7 +14,7 @@ branches:
 install:
   - pip install --upgrade numpy cython
   - pip install -e .
-
+  - pip install pytest pytest-cov
 
 script:
-  - true
+  - pytest --cov=./
```
kain88-de: `pytest --cov mdbenchmark -v --pep8 mdbenchmark`

this means you also need to install `pytest-pep8`

mimischi:
Should I squash some commits before merging?

.travis.yml
```diff
@@ -14,7 +14,7 @@ branches:
 install:
   - pip install --upgrade numpy cython
   - pip install -e .
-
+  - pip install pytest pytest-cov pytest-pep8
 
 script:
-  - true
+  - pytest --cov mdbenchmark -v --pep8 mdbenchmark
```
kain88-de: there still needs to be a 

```yaml
install:
   - pip install codecov

after_success:
    - codecov
```

kain88-de:
Yeah squashing would be nice.

kain88-de:
once we submit data to codecov you can also add a coverage token to the readme.

kain88-de:
oh and update the CHANGELOG

kain88-de:
@mimischi you can squash and merge when the tests pass

mimischi:
Just updated the CHANGELOG, bumped the version to `3.1.0` and squashed some commits. Looks like Codecov is having some hickups, so we may need to wait for it to finally finish the report.

I didn't write any tests for `utils.guess_ncores`, as I don't see how that should be that in a sensible way..?

I'm fine merging this PR.

kain88-de:
@mimischi no need to bump the version. We can actually accumulate thing between releases.

kain88-de:
I removed the version bump!

I will see on sunday how to publish a release on pypi and add some docs and tools to do that. I also want to experiment with `towncrier` to generate the changelog before we release.

mimischi:
Sounds good. I'll later add another test for the missing method in `utils.py` and then merge, when Codecov finishes.

kain88-de:
you can merge without codecov. That's OK as long as travis passes

PR #27, opened 2018-01-30
Crashfix and small usability update

### LINKED ISSUE #26: Analyze doesn't work if some of the runs did not complete
## Error report

**Expected behavior**

`mdbenchmark analyze` should show a performance of 0 or some other value (e.g. "simulation not finished"), if it can't find the `Performance` line in the output file.

**Actual behavior**

It exits due to [this line](https://github.com/bio-phys/MDBenchmark/blob/master/mdbenchmark/mdengines/gromacs.py#L49) and the user is shown the typical python stack trace, caused by the `warnings.warn`.

**Currently version of MDBenchmark:**
```bash
$ mdbenchmark --version
mdbenchmark, version 1.1.1
```

## Proposal

We previously just showed a default value of "0" for all the performances that were not found. We most definitely should go ahead and warn the user that specific systems have not finished yet/correctly, but the current implementation only works if all the runs have finished successfully.

I noticed when GROMACS was not able to do domain decomposition on more than 1 node, so these benchmarks crashed right after starting. The benchmark with one node did work, but `mdbenchmark` didn't show the actual results.

mimischi:
Fixes #26.

 - Changed the way that `mdengines/gromacs.py` warns the user. We now use `click.echo`, instead of `warnings.warn`. The former does not produce a stack trace in the console and is more customizable.
- Fixed three major bugs during `mdbenchmark analyze`
  1. Crash if log file was not present at all (simulation not started yet)
  2. Crash if performance was not found (simulation not started or still running)
  3. Crash if number of cores was not found (simulation not started yet)
- We now inform the user if some of the data (e.g. performance / number of cores) was not found and print it to the console.

This is sample output from a new data folder in `mdbenchmark/tests/data/analyze-files-w-errors`. Simulations 0-4 are complete ones, which we expect to work. In folder No. 5 we do not have a `.log` file. In number 6 the performance line is missing and in number 7 we deleted the line with the number of cores. I added a integration test to test for this exact output.

```
$ mdbenchmark analyze
WARNING We were not able to gather informations for all systems.
	Systems marked with question marks have either crashed or
	were not started yet.
          gromacs  nodes   ns/day  run time [min]    gpu   host ncores
0  gromacs/2016.3      1   98.147              15  False  draco     32
1  gromacs/2016.3      2  178.044              15  False  draco     64
2  gromacs/2016.3      3  226.108              15  False  draco     96
3  gromacs/2016.3      4  246.973              15  False  draco    128
4  gromacs/2016.3      5  254.266              15  False  draco    160
5  gromacs/2016.3      6        ?              15  False  draco      ?
6  gromacs/2016.3      7        ?              15  False  draco    160
7  gromacs/2016.3      8  254.266              15  False  draco      ?
```

### Todo

I'd like to do some other smaller stuff, but this PR may also get merged before. We finally need to do more unit and integration testing in this package.

- [x] Squash some of the commits
- [x] Add unit tests for the individual functions from `mdengines/gromacs.py`.

PR Checklist
------------
 - [ ] CHANGELOG updated (added [news fragment](https://github.com/hawkowl/towncrier#news-fragments) into changelog directory)?


mdbenchmark/mdengines/gromacs.py
```diff
@@ -46,8 +49,8 @@ def parse_ns_day(fh):
     for line in lines:
         if 'Performance' in line:
             return float(line.split()[1])
-    warnings.warn(UserWarning, "No Performance data found.")
-    return 0
+
+    return 0.0
```
kain88-de: why did you change this? The normal 0 worked just fine.

mdbenchmark/mdengines/gromacs.py
```diff
@@ -101,5 +104,16 @@ def analyze_run(sim):
     else:
         module = sim.categories['version']
 
+    # We were unable to find the information in the log file or the log file
+    # does not exist (yet). Append to the correct list in the provided
+    # dictionary.
+    if not ns_day or not ncores:
+        if not bool(ns_day):
```
kain88-de: Why not `if ns_day == 0` or `if np.allclose(ns_day, 0)`?

mdbenchmark/analyze.py
```diff
@@ -127,13 +135,25 @@ def analyze(directory, plot, ncores):
         'gromacs', 'nodes', 'ns/day', 'run time [min]', 'gpu', 'host', 'ncores'
     ])
 
+    errors = {'performance': [], 'ncores': []}
+
     for i, sim in enumerate(bundle):
-        df.loc[i] = analyze_run(sim)
+        df.loc[i], errors = analyze_run(sim, errors)
+
+    if errors['performance'] or errors['ncores']:
```
kain88-de: Instead of checking for errors why not just check if you have `NaN` values in the dataframe? This also avoids having to add the `error` parameter to `analyze_run`.

mdbenchmark/mdengines/gromacs.py
mimischi: The docstring says it returns a float, so I figured we should stick to it and also return a float in the default case.

mdbenchmark/mdengines/gromacs.py
mimischi: I changed `ns_day` and `ncores` to equal `False` on [lines 87-88](https://github.com/bio-phys/MDBenchmark/pull/27/files/9ae51acb59bbac121427b2fc3eb143acab1dfff8#diff-3a3de48ca1e8f5d2cc919ebde57cef2dR87-R88).

I could actually change it back again to equal `0` as a default, just thought `if not ns_day` is nicer in reading than `if ns_day == 0`.

mdbenchmark/analyze.py
mimischi: That's a good idea. I just changed missing values to `NaN` recently, after implementing the error collection. Getting rid of that is nicer.

mdbenchmark/mdengines/gromacs.py
kain88-de: Actually, you can just use `np.nan` as a default and also return that value in case we cannot read any information. This is sufficient to mark the error. Then there is no need for additional error handling in this function. Your error handling idea of using `nan` in the analysis function is pretty good and by returning nan we make better use of it.

mdbenchmark/mdengines/gromacs.py
kain88-de: Returning `0` or `0.0` is equivalent in python2/3. The type is implicitly upcast in a operation with a float. Only in python 2 the division by an integer is problematic since that returns an int. But 0 divided by anything is also 0. So for 0, it doesn't make a difference.

kain88-de: But as I have written above returning a `np.nan` is actually a better decision.

mimischi:
> Could you please also minimize the new example data. It's quite long right now.

Do you mean cutting down the file sizes or only minimizing the number of systems?

kain88-de:
cutting down filesize. We add 100k lines in this PR.  I admit I haven't been good about this either. I should shrink the files we have right now as well.

mimischi:
I can just do that for all files if you're fine with it!

kain88-de:
Sure go ahead.

Michael Gecht writes:

mimischi:
Done. Just used `for FILE in $(ls */*/bench.log -1); do (head -n100 && tail -n5) <$FILE > "$FILE.short"; mv $FILE.short $FILE -f; done` inside of `mdbenchmark/tests/data`. This kept the first 100 and last 5 lines of each log file.

mimischi:
@kain88-de I just added some unit tests to the functions `gromacs.parse_ns_day` and `gromacs.parse_ncores`. I'd like to do the same thing with `gromacs.analyze_run`, but I can't access the `tests/data` directory with the pytest fixtures you wrote in `testing.py`. They have that hardcoded relative path, assuming all our test files are located in `tests/`, but I think it's better to use a folder structure for tests similar to the actual module.

Could you help me out on how to solve this?

Also codecov seems to have some hiccups, as it did not show any coverage reports yet.

Issue #24, opened 2018-01-19
Extend Plotting options


kain88-de:
Our plotting option is limited right now. It would be nice if we can extend to allow plotting with several gromacs versions.  The general roadmap for this would be.

- [x] factor out plotting from `analyze`
- [x] have function that can add a line plot for one benchmark run (gromacs version/host/gpu)
- [x] maybe have a new command `mdbenchmark plot` that takes a csv from as input.
- ~[ ] suggest which number of nodes have the best performance~

MSiggel:
the same principle should also be implemented to differentiate different md engines like namd and gromacs for sake of easy comparisons.

kain88-de:
The interface of the function to plot a single benchmark (same module, same system, etc) should have the following API. 

```python
def plot_single_benchmark(..., ax=None):
    if ax is None:
        ax = plt.gca()
    ...
    return ax
```

This gives convenience for plotting as a library as well as full control if someone wants to create more complicated plots.

kain88-de:
The plotting options should probably take an argument `groupby` that takes a list of variables that we want to separate for the plotting. So that e.g.

```bash
mdbenchmark plot results.csv --groupby="gpu, module, host"
```
Will produce a line for each unique combination of values in the `gpu`, `module` and `host` columns.
The current default is then `--groupby="gpu"`.  This allows to be flexible in the allowed graphs that can be produced with a minimal UI. Your code should catch column names like `nodes` or `ns/day` to be invalid for a groupby option because those are used for the x and y axis.

In the code this should look similar to

```python
cols = groupby.split(',')
# This should generate an iterable of dataframes for each unique combination of
# the provided cols.
runs = dataframe.groupby(cols)
for df in runs:
    plot_single_benchmark(df, ax=ax, ...)

```

This means we support and an arbitrary number of lines to the plot. This, of course, opens the question about colors. We can actually let matplotlib deal with that.

```python
# plot measured results
p = ax.plot(x, y)
# obtain last used color object.
c = p[0].get_color()
# plot performance fit in same color
ax.plot(x, y, color=c)
```

mimischi:
As a note, so we do not forget about this:

If for whatever reasons performance from some nodes is not available (skipped benchmarks with `n`-nodes, benchmark has not yet finished, ...), the plotting is not 100% reliable. Using the current `mdbenchmark version 1.2.0` release, this happens:

```bash
mdbenchmark analyze
WARNING	We were not able to gather informations for all systems.
[...]

PR #29, opened 2018-02-01
Add NAMD benchmark functionality

MSiggel:
## Work in Progress
- added automated md engine detection to untils
- analysis and generate both use the respective engine file in mdengines to generate and analyze data for the respective engine
- mdengine file added for namd
- updated draco/hydra templates for namd 
- added separate plot function in plot.py (devlopment stage)
- preliminary changes for analyze.py (development stage)


MSiggel:
work in progress:
- test files have to be adapted for the new changes
- analyze and plot are not finished and need to be debugged afterwards
- discussed API structure of plotting functions not yet implemented
- mdengine files can be consolidated and shortened significantly

mdbenchmark/plot.py
```diff
@@ -0,0 +1,273 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
```
kain88-de: could you please move this into a separate branch.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,185 @@
...
+    -------
+    float
+        number of cores job was run on
+    """
+    if isinstance(fh, string_types):
+        with open(fh) as f:
+            lines = f.readlines()
+    else:
+        lines = fh.readlines()
+        fh.seek(0)
```
kain88-de: we simplified this now to

```python
lines = fh.readlines()
```
 in master an just always assume that `fh` is a filehandle.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,185 @@
...
+    Analyze Performance data of a NAMD simulation
+    """
+    ns_day = 0
+
+    # search all output files and ignore GROMACS backup files
+    output_files = glob(os.path.join(sim.relpath, '[!#]*log*'))
+    if output_files:
+        with open(output_files[0]) as fh:
+            ns_day = parse_ns_day_namd(fh)
+            ncores = parse_ncores_namd(fh)
```
kain88-de: this means here you need to reset the filehandle with `fh.seek(0)` between calls.

mdbenchmark/mdengines/gromacs.py
```diff
@@ -103,3 +108,49 @@ def analyze_run(sim):
...
+        module=module,
+        formatted_module=formatted_module,
+        n_nodes=nodes,
+        time=time,
+        formatted_time=formatted_time)
+
+    with open(sim['bench.job'].relpath, 'w') as fh:
+        fh.write(script)
+
+def formatted_md_engine_name(modulename):
```
kain88-de: duplicate

mdbenchmark/mdengines/gromacs.py
```diff
@@ -103,3 +108,49 @@ def analyze_run(sim):
...
+
+    if not os.path.exists(tpr):
+        raise click.FileError(
+            tpr, hint='File does not exist or is not readable.')
+
+    copyfile(tpr, sim[tpr].relpath)
+    # Add some time buffer to the requested time. Otherwise the queuing system
+    # kills the jobs before GROMACS can finish
+    formatted_time = '{:02d}:{:02d}:00'.format(*divmod(time + 5, 60))
+    formatted_module = "gromacs"
```
kain88-de: a better name is `mdengine`

mdbenchmark/generate.py
```diff
@@ -140,7 +116,7 @@ def generate(name, gpu, module, host, max_nodes, min_nodes, time, list_hosts):
             gromacs_module, gpu_string))
```
kain88-de: gromacs module seems to be wrong.

MSiggel:
This version contains works with some toy files which are included in the sample files. One can generate gromacs and namd  benchmarks with generate once if the names of all the required files are named the same. Analyze can read out namd or gromacs file in one directory. 
Testing on cluster still pending..

mdbenchmark/utils.py
```diff
@@ -125,3 +128,24 @@ def guess_ncores():
...
+    else:
+        raise RuntimeError("No Module Detected! did you specify the module?")
+
+def formatted_md_engine_name(modulename):
+    if 'gromacs' in modulename:
+        return "gromacs"
+    elif 'namd' in modulename:
+        return "namd"
+    else:
+        raise RuntimeError("No Module Detected! did you specify the module?")
```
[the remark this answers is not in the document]
kain88-de: To complicated. 

In gromacs.py / namd.py
```python
import ...

NAME = "GROMACS"

def ...
```

Any function that later wants to name can do.

```python
get_md_engine(module).NAME
```

The idea is that every module carries a formatted name. This means we don't need to distinguish between the case where we want a name or a module.

mdbenchmark/utils.py
```diff
@@ -125,3 +128,24 @@ def guess_ncores():
...
+    """Detects the MD engine based on the available modules.
+        Any newly implemeted mdengines must be added here.
+    """
+    if 'gromacs' in modulename:
+        print(modulename)
+        return mdengines.gromacs
+    elif 'namd' in modulename:
+        return mdengines.namd
+    else:
+        raise RuntimeError("No Module Detected! did you specify the module?")
```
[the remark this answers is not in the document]
kain88-de: Maybe we can add our own `warning` and `critical` wrappers.

```python
def warning(msg):
    click.echo("WARNING: " + msg, ...)

def critical(msg):
    click.echo("CRITICAL ERROR: " msg, ...)
    sys.exit(1)
```

That would make consistent formatting easier.

mdbenchmark/utils.py
```diff
@@ -125,3 +128,24 @@ def guess_ncores():
...
         total_cores = 1
     return total_cores
+
+#detects the module names and returns the package
+def detect_md_engine(modulename):
+    """Detects the MD engine based on the available modules.
+        Any newly implemeted mdengines must be added here.
+    """
+    if 'gromacs' in modulename:
+        print(modulename)
```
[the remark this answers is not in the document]
MSiggel: yeah this is an old test..

mdbenchmark/utils.py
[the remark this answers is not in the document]
MSiggel: formatted_md_engine_name is not called anymore so it can be removed. as soon as the module was detected the respective engine files "know" their formatted and they can be hard coded.

mdbenchmark/templates/hydra
```diff
@@ -19,4 +19,8 @@ module purge
 module load {{ module }}
 
 # run {{ module }} for {{ time }} minutes
+{%- if formatted_module is gromacs %}
```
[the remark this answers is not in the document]
MSiggel: ah yes.. nevermind, I forgot to change that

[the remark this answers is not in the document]
mimischi: Lets maybe stick to one, just in case users get confused when comparing different templates :)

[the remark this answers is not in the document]
MSiggel: yes, this was faulty syntax anyway

mimischi:
@MSiggel Tell me when you are ready, then I can perform a rebase onto `master` for you.

mdbenchmark/utils.py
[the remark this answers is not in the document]
MSiggel: @kain88-de, @mimischi: I think this would be a good way to stay consistent!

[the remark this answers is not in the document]
mimischi: @MSiggel Should we create a new PR for this and merge it into `master`, so you can adapt it in here?

MSiggel:
this should be all @mimischi. I haven't introduced any changes to the errors yet until we introduce our own wrappers for these.

mdbenchmark/analyze.py
```diff
@@ -131,15 +132,38 @@ def plot_analysis(df, ncores):
     help='Number of cores per node. If not given it will be parsed from the '
     'benchmarks log file.',
     show_default=True)
-def analyze(directory, plot, ncores):
+@click.option(
+    '-e',
```
kain88-de: I think this should be removed again. But in general, don't add too many short options. Only for options that are set often and that can be set to have the same letter as the start of the long version.

mdbenchmark/generate.py
```diff
@@ -134,18 +115,29 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
+        # here we switch between the md engines like gromacs and namd
+        # namd input files can have relative imports. To avoid issues when
+        # copying the files we currently only allow the use of absolute paths. 
+        engine = detect_md_engine(m)
+        if engine == namd:
+            click.secho('!!!ATTENTION!!!', bold=True)
+            click.echo("""This is an experimental NAMD support!
+All files must be in this directory.
+Parameter paths must be abslute in NAMD files!
+            """)
```
kain88-de: Maybe rather add this into the generate function of the NAMD module.

mdbenchmark/mdengines/gromacs.py
```diff
@@ -17,9 +17,18 @@
...
 import os
+import warnings
 from glob import glob
+from shutil import copyfile
+
+import click
+import datreant.core as dtr
+import mdsynthesis as mds
 import numpy as np
+from jinja2.exceptions import TemplateNotFound
```
kain88-de: Do you specifically catch this exception somewhere? If not I would remove the import.

mdbenchmark/utils.py
```diff
@@ -125,3 +127,17 @@ def guess_ncores():
                        click.style('WARNING', fg='yellow', bold=True)))
         total_cores = 1
     return total_cores
+
+
+
+def detect_md_engine(modulename):
```
kain88-de: I talked with @mimischi today. This would be better suited in the mdengine/__init__.py file.

mdbenchmark/utils.py
```diff
@@ -125,3 +127,17 @@ def guess_ncores():
...
+    """Detects the MD engine based on the available modules.
+        Any newly implemeted mdengines must be added here.
+        Returns the python module.
+    """
+    if 'gromacs' in modulename:
+        return gromacs
+    elif 'namd' in modulename:
+        return namd
+    else:
+        raise RuntimeError("No Module Detected! did you specify the module?")
```
kain88-de: Please write in the error message that the mdengine couldn't be detected from the module and what mdengines are known to mdbenchmark.

mdbenchmark/generate.py
MSiggel: if one moves it to the namd generate function it will be shown x times for each folder which is generated..maybe a new util fxn? we'll probably need such a warning for other engines too.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def analyze_run(sim):
+    """
+    Analyze Performance data of a NAMD simulation
+    """
+    ns_day = np.nan
+    ncores = np.nan
+
+    # search all output files and ignore GROMACS backup files
```
mimischi: This comment still says `GROMACS`.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+    output_files = glob(os.path.join(sim.relpath, '[!#]*out*'))
+    if output_files:
+        with open(output_files[0]) as fh:
+            ns_day = parse_ns_day(fh)
+            fh.seek(0)
+            ncores = parse_ncores(fh)
+
+    # Backward compatibility to previously created mdbenchmark systems
```
mimischi: We probably do not need this here. We added it to the GROMACS module, because we added more categories to your `Sim` objects at some point. All new Sim objects for NAMD will include that category `time`.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+    if 'time' not in sim.categories:
+        sim.categories['time'] = 0
+
+    # backward compatibility
```
mimischi: Same thing here. I do not think we will need this check. Just using `sim.categories['module']` for NAMD should be fine. What are your thoughts @kain88-de?

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+        module = sim.categories['version']
+
+    return (module, sim.categories['nodes'], ns_day, sim.categories['time'],
+            sim.categories['gpu'], sim.categories['host'], ncores)
+
+
+def write_bench(top, tmpl, nodes, gpu, module, tpr, name, host, time):
+    """ Writes a single namd benchmark file and the respective sim object
+    """
+    # TODO: do this centrally somewhere not here! needed in any file
```
mimischi: What about this TODO?

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+        if not os.path.exists(filename):
+            raise click.FileError(
+                filename, hint='File does not exist or is not readable.')
+
+    copyfile(namd, sim[namd].relpath)
+    copyfile(psf, sim[psf].relpath)
+    copyfile(pdb, sim[pdb].relpath)
+
+    # Add some time buffer to the requested time. Otherwise the queuing system
+    # kills the jobs before GROMACS can finish
```
mimischi: Comment says `GROMACS`.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+        formatted_module=md_engine,
+        n_nodes=nodes,
+        time=time,
+        formatted_time=formatted_time)
+
+    with open(sim['bench.job'].relpath, 'w') as fh:
+        fh.write(script)
+
+
+def analyze_namd_file(fh):
```
mimischi: What is this function for?

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,165 @@
...
+    """ Check for parameter files in the namd config file and copy them here
+    """
+
+    return 1
+
+
+def parse_namd_file():
```
mimischi: What is this function for?

mdbenchmark/templates/draco
```diff
@@ -29,7 +29,12 @@
...
 module purge
 module load impi
+module load cuda
 module load {{ module }}
 
 # Run {{ module }} for {{ time  }} minutes
+{%- if formatted_module == "gromacs" %}
 srun gmx_mpi mdrun -v  -maxh {{ time / 60 }} -deffnm {{ name }}
+{%- elif formatted_module == "namd" %}
+srun -np {{24 * n_nodes}} namd2 {{ name }}
```
mimischi: Should we also add the `-np` option for the `gmx_mpi` call?

mdbenchmark/tests/test_generate.py
```diff
@@ -43,6 +43,7 @@ def test_generate(cli_runner, tmpdir, tpr_file):
             'generate', '--module=gromacs/2016', '--host=draco',
             '--max-nodes=4', '--gpu', '--name={}'.format(tpr_file)
         ])
+        print(result)
```
mimischi: This seems to be obsolete

mdbenchmark/analyze.py
```diff
@@ -132,15 +133,31 @@ def plot_analysis(df, ncores):
     help='Number of cores per node. If not given it will be parsed from the '
     'benchmarks log file.',
     show_default=True)
-def analyze(directory, plot, ncores):
+@click.option(
+    '-o',
+    '--output_name',
+    help="Name of the output .csv file.",
+    default="runtime.csv",
```
mimischi: Do we want to keep it this way? I initially thought we will use the `host`, `module` and `gpu` values for the output file name? This way we could prepare for #24, where we would specify different columns to group by. These be used for the filenames..? Or am I overthinking this?

mdbenchmark/generate.py
```diff
@@ -26,53 +26,32 @@
...
-        name=name,
-        gpu=gpu,
-        module=module,
-        n_nodes=nodes,
-        time=time,
-        formatted_time=formatted_time)
-
-    with open(sim['bench.job'].relpath, 'w') as fh:
-        fh.write(script)
+#import mdengines
```
mimischi: This is probably obsolete

mdbenchmark/generate.py
```diff
@@ -26,53 +26,32 @@
...
 
 
 @cli.command()
 @click.option(
-    '-n', '--name', help='Name of .tpr file.', default='md', show_default=True)
+    '-n',
+    '--name',
+    help='Name of input files. All files must have the same base name.',
+    default='',
```
mimischi: Do we need `default=''` here? We can probably get rid of it

mdbenchmark/generate.py
```diff
@@ -123,6 +102,7 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
         raise click.BadParameter(
             'You did not specify which gromacs module to use for scaling.',
             param_hint='"-m" / "--module"')
+    #detect the engine and use accordingly
```
mimischi: This comment throws an pep8 error

mdbenchmark/generate.py
mimischi: Will we keep this message when merging? We should not merge anything that is not _fully functional_? If we are sure that it is working, which the unit/integration test _should_ pledge for, then we can get rid of that message entirely. Also trying it out on the clusters would help reassure that everything is working :fire:

mdbenchmark/mdengines/namd.py
kain88-de: yeah this is only needed for gromacs

mdbenchmark/analyze.py
kain88-de: you are overthinking this.

mdbenchmark/generate.py
kain88-de: yeah, this shouldn't have a default.

kain88-de:
> Will we keep this message when merging? We should not merge anything that is not fully functional? If we are sure that it is working, which the unit/integration test should pledge for, then we can get rid of that message entirely. Also trying it out on the clusters would help reassure that everything is working

Yes, we will keep it. NAMD isn't as foolproof to use in MDBenchmark as GROMACS is. NAMD is reading a lot of files when it's started and we make the assumption that all file paths are specified as absolute paths. Making this work for *all* NAMD inputs might take a little bit. Therefore the warnings.

MSiggel:
I moved the tpr check to a new function in the gromacs file as this is gromacs specific. I think this improves modularity. Additionally, for namd we need to check for other files anyway. Moreover, the order of checks changed because the check function needs the engine.
While doing so I also realized that the generate function still uses a tpr argument which should be dropped or renamed as this is no longer valid for namd use.

mdbenchmark/mdengines/namd.py
MSiggel: for now i would leave this. Once namd is merged we can think about how we manage this for all engines i guess?

MSiggel:
@mimischi:  I added the requested unique file name (based on date and time) generator to analyze. 
Additionally, I added a crude NAMD file check which looks at parameter and input file paths and infers whether an absolute path was given...This is probably the minimum we should do for the user.

mdbenchmark/tests/data/analyze-files-namd/2/bench.job
```diff
@@ -0,0 +1,24 @@
...
+#SBATCH --ntasks-per-node=32
+# Wall clock limit:
+#SBATCH --time=00:20:00
+
+module purge
+module load impi
+module load namd
+
+# Run namd for 15 minutes
+namd2 apoa1
```
mimischi: This is very minor, but the syntax for the `namd2 apoa1` call is not the one that we put into `templates/draco`. Maybe we should put them to the same state :)

mdbenchmark/generate.py
```diff
@@ -119,22 +89,24 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
     for m in module:
         directory = '{}_{}'.format(host, m)
         gpu_string = '.'
 
+        # Here we detect the mdengine (GROMACS or NAMD).
+        # Note: NAMD input files can have relative imports. To avoid issues
+        # when copying the files, we currently only allow the use of absolute
+        # paths.
+        engine = detect_md_engine(m)
+        if engine == namd:
```
mimischi: Could you please put this into one single `console.warn` call? Also we probably can skip the first line and only use something like this: `console.warn('NAMD support is experimental. All input files must be in the current directory. Parameter paths must be absolute. Only crude file checks are performed!')`

kain88-de:
I did some cleanup and rebased against master. The function to clean up a previous simulation is not module dependent, to work for NAMD as well. I also based the function on a whitelist of allowed files instead of a blacklist.

mdbenchmark/generate.py
```diff
@@ -124,33 +89,39 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
     for m in module:
-        directory = '{}_{}'.format(host, m)
-        gpu_string = '.'
+        # Here we detect the mdengine (GROMACS or NAMD).
+        engine = detect_md_engine(m)
+        # Note: NAMD input files can have relative imports. To avoid issues
+        # when copying the files, we currently only allow the use of absolute
+        # paths.
+        if engine == namd:
+            console.warn('ATTENTION NAMD USERS!'
```
mimischi: We probably can skip the first line (`ATTENTION` is not needed when the message is prefixed with a bold and colored `WARNING`).  Use something like this instead: console.warn('NAMD support is experimental. All input files must be in the current directory. Parameter paths must be absolute. Only crude file checks are performed!')

kain88-de:
The example in the readme needs to be updated! Maybe link to the hydra template directly that might be possible in RST and would allow that any change in the template is immediately rendered in the readme.

Also, an interesting question would this update break existing user templates?

mimischi:
I just added some more commits. So far I'm pretty happy with this and I think we can start squashing commits, I could also do that.

There is one thing left that I do not like. If one supplies a file name that does not exist (in this case `md`), then our script first tells the user it will create benchmarks and _then_ fail, because it cannot find files. I would rather like the program to _first_ check if all files are present and fail immediately if this is not the case.

```bash
$ mdbenchmark generate --host draco --module namd --min-nodes 4 --max-nodes 6 -n md
WARNING NAMD support is experimental. All input files must be in the current directory. Parameter paths must be absolute. Only crude file checks are performed!
Creating benchmark system for namd.
Creating a total of 3 benchmarks, with a run time of 15 minutes each.
ERROR File 'md.namd' for benchmark does not exists.
```

mimischi:
Just found some more bugs. Working on it.

mimischi:
Tests should be fine now. Regarding my above comment https://github.com/bio-phys/MDBenchmark/pull/29#issuecomment-368296461, I changed two things:

1. Check that the required files are present _before_ we print anything else to the user. This will only work when using the same engine. Using intermixed engines (namd and gromacs), we will check files for the first engine and print things to the console. Then we will check files for the next engine and _could_ fail at this point, if these files are not present.
2. Move the `NAMD experimental` warning outside of the `for m in module` loop. This way, we only print that warning once at the beginning.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def analyze_namd_file(fh):
+    """ Check whether the NAMD config file has any relative imports or variables
+    """
+    lines = fh.readlines()
+
+    for line in lines:
+        # Continue if we do not need to do anything with the current line
+        if ('parameters' not in line) and ('coordinates' not in line) and (
```
MSiggel: these options always have to be provided otherwise the namd run will completely fail. Therefore, this wouldn't serve any purpose. If these are not found we should exit and trown an error or something.

mdbenchmark/mdengines/namd.py
```diff
@@ -0,0 +1,196 @@
...
```
mimischi: @MSiggel I could not find your comment anymore. I just simplified your previous code, which repeated the `console.error` lines for all three cases.

If you are saying that we need to change the structure in general, then we should do this. Your previous code did not do it either.

mimischi:
I left two `TODO` entries with the `cleanup` functions in both mdengines. Somehow I cannot get the white_list to keep `bench.log`. I excluded these from the tests, as they were never inside of them before.

mdbenchmark/generate.py
```diff
@@ -71,6 +71,11 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
         print_possible_hosts()
         return
 
+    if not name:
```
kain88-de: isn't there a click option to print this automatically?

mdbenchmark/tests/mdengines/test_gromacs.py
```diff
@@ -86,7 +86,7 @@ def test_cleanup_before_restart(tmpdir):
     """Test that the cleanup of each directory works as intended."""
     FILES_TO_DELETE = [
         'job_thing.err.123job', 'job_thing.out.123job', 'md.log', 'md.xtc',
-        'md.cpt', 'md.edr', 'job.po12345', 'job.o12345', 'md.out'
+        'md.cpt', 'md.edr', 'job.po12345', 'job.o12345', 'md.out', 'bench.job'
```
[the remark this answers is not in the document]
mimischi: We would only call the cleanup when restarting, so it would be best to get rid of the log for the previous run.

mdbenchmark/generate.py
mimischi: Only if we define it as an argument, then we could use `click.File` inside of that definition.

mdbenchmark/tests/mdengines/test_gromacs.py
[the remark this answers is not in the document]
kain88-de: oh sorry I meant bench.**job** the submit script.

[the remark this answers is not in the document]
mimischi: Yeah, I only put it there so the test would pass. Could have used `@pytest.mark.skip` for that.

mimischi:
I squashed all 88 commits into only two: one doing all the heavy lifting on the main code and the other for the tests.

mimischi:
@kain88-de This should be finished now. Should we update the README/AUTHORS here or in a separate PR after merging?

mdbenchmark/analyze.py
```diff
@@ -132,15 +133,27 @@ def plot_analysis(df, ncores):
...
-def analyze(directory, plot, ncores):
+@click.option(
+    '-o', '--output-name', default=None, help="Name of the output .csv file.", type=str)
+def analyze(directory, plot, ncores, output_name):
     """Analyze finished benchmarks."""
     bundle = mds.discover(directory)
+
     df = pd.DataFrame(columns=[
-        'gromacs', 'nodes', 'ns/day', 'run time [min]', 'gpu', 'host', 'ncores'
+        'module', 'nodes', 'ns/day', 'run time [min]', 'gpu', 'host', 'ncores'
```
kain88-de: If we want to stay backward compatible we should keep the `gromacs` column. Any other module then gromacs would be stored with a '?' in this columns. The `module` column should still be added. This would ensure that plotting scripts in the wild based on the old files continue to work.

mdbenchmark/mdengines/gromacs.py
```diff
@@ -96,3 +103,75 @@ def analyze_run(sim):
...
+    if not os.path.exists(tpr):
+        console.error(
+            "File {} does not exist, but is needed for GROMACS benchmarks.",
+            tpr)
+
+    return tpr
+
+
+def cleanup_before_restart(sim):
+    white_list = ['((?:[^/]*/)*)(bench\.job)', '.*.tpr', '.*.mdp']
```
kain88-de: simpler regex is better.

mimischi:
@MSiggel Could you update the `README.rst` explaining how to use the NAMD feature? This would help users get started quickly. If this is done, we can merge! :fire:

MSiggel:
I just realized there is a remaining issue: NAMD has different builds for with or without GPU (usually _cuda). One would assume that the user knows this and uses the right module on the cluster but theoretically this might be a cause of problems for inexperienced users.. Do we need to catch this? NAMD will still run but the benchmark times will be wrong.

mimischi:
That's an excellent point. It looks like there is only the normal NAMD version on `draco`, while there is also `namd/version-gpu` on `hydra`.

I would suggest the following: put a notice into the README making the users aware of this. We can implement this restriction into #49, after this gets merged.

kain88-de:
Since we already have the NAMD warning message. I would suggest as a first step to add an info/warning box here that the user should ensure the module is GPU compatible when he has selected the `--gpu` switch.

README.rst
```diff
@@ -120,16 +123,77 @@ You can also create benchmarks for different versions of GROMACS:
...
 
-As soon as the benchmark simulations have been submitted you can run the
-analysis script via ``mdbenchmark analysis``. As soon as at least one system has
-finished the script will produce a ``.csv`` output file or a plot for direct
-usage (via the ``--plot`` option).
+To start specific benchmarks separately, use the `--directory` option and specify the corresponding folder:
+
+.. code::
+
+    $ mdbenchmark start --directory draco_gromacs/5.1.4-plumed2.3
```
kain88-de: use submit. start gives a deprecation warning.

README.rst
```diff
@@ -120,16 +123,77 @@ You can also create benchmarks for different versions of GROMACS:
...
+-----------------------------
+
+**NAMD support is experimental. If you encounter any problems or bugs, we would
+appreciate to `hear from you`_.**
+
+Generating benchmarks for NAMD follows a similar process to GROMACS. Assuming
+the NAMD configuration file is called ``protein.namd``, you will need the
+corresponding ``protein.pdb`` and ``protein.psf`` files to generate benchmarks.
+**Please be aware that all paths given in the ``protein.namd`` file must be
+absolute paths**. This makes sure that MDBenchmark does not destroy paths when
```
kain88-de: this ensures that

mimischi:
This looks great. Thanks for the effor @MSiggel!

PR #36, opened 2018-02-13
Update project internals

mimischi:
- Added `Makefile` to project.
- Updated help text and docstrings for `generate`, `submit` and `analyze`.
- Rephrased and formatted messages printed to the console.
- `mdbenchmark generate` will now _first_ look for the `.tpr` file and only then do everything else.
- Added unit tests for all messages generated via `click.echo()` and `click.BadParameter()`.
- Add unit test for `guess_ncores()` utility function.

PR Checklist
------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+.PHONY: build clean clean-build clean-pyc clean-test upload
+
+clean: clean-build clean-pyc clean-test
+
+clean-build:
+	rm -rf build/
+	rm -rf dist/
+
+clean-pyc:
+	find . -name '*.pyc' -exec rm -f {} +
+	find . -name '*.pyo' -exec rm -f {} +
+	find . -name '*~' -exec rm -f {} +
+	find . -name '__pycache__' -exec rm -rf {} +
+
+clean-test:
+	rm -rf .tox/
+	rm -f .coverage
+	rm -rf htmlcov/
+	rm -rf .pytest_cache/
+
+build: clean
+	python setup.py sdist bdist_wheel --universal
+
+upload: build
+	twine upload dist/*
--- a/mdbenchmark/generate.py
+++ b/mdbenchmark/generate.py
@@ -29,7 +29,7 @@ from .cli import cli
 from .utils import ENV, normalize_host, print_possible_hosts
 
 
-def write_bench(top, tmpl, nodes, gpu, module, name, host, time):
+def write_bench(top, tmpl, nodes, gpu, module, tpr, name, host, time):
     sim = mds.Sim(
         top['{}/'.format(nodes)],
         categories={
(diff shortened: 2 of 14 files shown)


Makefile
```diff
@@ -0,0 +1,10 @@
+build:
+	python setup.py sdist bdist_wheel --universal
+
+clean-build:
+	rm --force --recursive build/
+	rm --force --recursive dist/
+	rm --force --recursive *.egg-info
```
kain88-de: this might delete your developer installation!

Makefile
```diff
@@ -0,0 +1,10 @@
...
+
+upload:
+  twine upload dist/*
```
kain88-de: add a dependency chain here. 


```Makefile
.PHONY clean-build build upload

clean-build: 

build: clean-build

upload: build
```

this way we can just say upload and are always sure that the build is clean.

mdbenchmark/generate.py
```diff
@@ -141,5 +141,6 @@ def generate(name, gpu, module, host, max_nodes, min_nodes, time, list_hosts):
         for n in range(min_nodes, max_nodes + 1):
             write_bench(top, tmpl, n, gpu, m, name, host, time)
 
-    click.echo('Finished generating all benchmark systems.')
-    click.echo('Now run `mdbenchmark start` to submit jobs.')
+    click.echo('Finished generating all benchmarks.')
+    click.echo('You can now submit the jobs with {}.'.format(
+        click.style('mdbenchmark start', bold=True)))
```
kain88-de: benchmark submit

mimischi:
@kain88-de Besides the missing changelog fragments, this PR should be finished. I have got two questions:

1. All code in `mdbenchmark/ext` is untested. Should we write some small tests for this or just tell `codecov` to ignore that folder? These files are diminishing our overall code coverage.
2. I changed the exit codes to `sys.exit(1)`, when we exit with an error. I noticed that there is also exit code `2`, referring to bad CLI usage. Should we use that when applicable?

mimischi:
Oh, I just noticed a weird edge case with `towncrier`: if we add several independent features/fix bugs in a PR, we actually need to create several `#PR.feature` or `#PR.bugfix` files, but obviously this is not possible.

Putting several changes on multiple lines does also not work in `towncrier`. What do we do in this case?

kain88-de:
> What do we do in this case?

Several options. 

1. Notice that you included too many changes in a single PR and split it up
2. Check the towncrier docs/issues for a solution
3. ditch towncrier.

kain88-de:
> All code in mdbenchmark/ext is untested.

It's not tested anymore in this repo. But the code has been tested before. We only use it as a backup now anyway. 

> I changed the exit codes to sys.exit(1), when we exit with an error. I noticed that there is also exit code 2, referring to bad CLI usage. Should we use that when applicable?

No just a 1 for error occured. There is no real consistency with exit codes in TUI applications. So let's just stick with 0-OK or 1-Error.

mimischi:
@kain88-de I can't find any other suitable replacement, nor does anyone comment on this issue in the `towncrier` repository. I would say we just keep this PR, as I just added two short changelog fragmnets.

If you are fine with the PR, we can merge and then I would finally prepare a new release (to finally have the bugfix for `mdbenchmark analyze` on `pip`/`conda`, see #27).

Issue #38, opened 2018-02-19
streamline errors and warnings


MSiggel:
As a new version will contain multiple engines and we'll probably have to throw more warnings or errors, it probably would be good to streamline the warnings and errors so they all look uniform. At the moment there are many outputs when I run generate and we should clearly highlight important information. 
@mimischi:  we've talked about this before..

mimischi:
That is an excellent point. In the latest PR #36 I updated all warnings and errors  thrown at the user. Nonetheless, we are not 100% consistent.

We use [`click.echo`](http://click.pocoo.org/6/utils/#printing-to-stdout) for custom messages. For example to inform the user that we are generating new benchmarks or that we have finished doing so. Some info is presented in bold/colored text, if the console supports that (e.g., `mdbenchmark submit` is bold in the last line in the following example). If the console does not support it, we do not have a clear distinction of the command. Maybe we should put single quotes around the commands? But this would look ugly if we have both single quotes and bold text fonts, although I would not want to get rid of the latter.

```bash
$ mdbenchmark generate --host hydra -m gromacs/2016.4
Creating a total of 5 benchmarks, with a run time of 15 minutes each.
Creating benchmark system for asd.
Finished generating all benchmarks.
You can now submit the jobs with mdbenchmark submit.
```

If prefixed with `ERROR` (bold, red) or `WARNING` (bold, orange), the user is informed of some usage error (.tpr file not found for GROMACS, no benchmarks found in the give path, ...) or a warning (not all benchmarks have finished and we therefore cannot output all performance data).

```bash
$ mdbenchmark analyze -d non_existent_path
ERROR There is no data for the given path.
```

Furter we use [`click.BadParameter`](http://click.pocoo.org/6/api/#click.BadParameter) , when the user did not use the CLI as intended (wrong command/missing parameters). The problem here is that `click.BadParameter` has a specific style to its error messages that differs to our custom style for errors and warnings:

```bash
$ mdbenchmark generate --host asd
Usage: mdbenchmark generate [OPTIONS]

Error: Invalid value for "--host": Could not find template for host 'asd'.
```

I would rather ditch `click.BadParameter` and use a custom `click.echo`, but I would like to hear other thoughts first.

mimischi:
Before I forget, currently `mdbenchmark analyze` is pretty non-verbose and sometimes spends several seconds without any output to the user. We should think about adding some spinner to the CLI, which would _at least_ give some feeling of something happening.

PR #42, opened 2018-02-20
Consolidate all console output into one API

mimischi:
Related to #38.

We now can change the format of all info, warning and error messages by only changing one specific function. Messages that should _never_ be formatted were left with `click.echo` (one instance, when listing available host templates).

- Added internal API to generalize the use of `click.echo`.
- Messages to the user are now displayed via `mdbenchmark.formatting.info`, `mdbenchmark.formatting.warn` and `mdbenchmark.formatting.error`.
- All styling is done in `formatting.py`. We _never_ call `click.style` anymore, as special styling (making text bold) can be applied with the `placeholder` argument. 

PR Checklist
------------


mdbenchmark/utils.py
```diff
@@ -57,7 +59,7 @@ def get_possible_hosts():
 
 def print_possible_hosts():
     all_hosts = get_possible_hosts()
-    click.echo('Available host templates:')
+    formatting.info('Available host templates:')
     for h in all_hosts:
         click.echo('{}'.format(h))
```
mimischi: I let this in on purpose, as this should always be printed without any further formatting (in case `formatting.info` changes in the future.)

kain88-de:
```python
import click
import six


def echo(message, prefix=None, args=None, kwargs=None):
    if args is None:
        args = ()
    if kwargs is None:
        kwargs = {}

    if prefix is not None:
        message = '{}: {}'.format(prefix, message)

    if args:
        args = [click.style(str(el), bold=True) for el in args]
    if kwargs:
        kwargs = {
            k: click.style(str(v), bold=True)
            for k, v in six.iteritems(kwargs)
        }
    click.echo(message.format(*args, **kwargs))

    
def error(message, *args, **kwargs):
    echo(message, prefix="ERROR", args=args, kwargs=kwargs)
```

mdbenchmark/generate.py
```diff
@@ -140,13 +141,13 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
+            'Creating benchmark system for {}', placeholders=m + gpu_string)
 
         for n in range(min_nodes, max_nodes + 1):
             write_bench(top, tmpl, n, gpu, m, tpr, name, host, time)
 
-    click.echo('Finished generating all benchmarks.')
-    click.echo('You can now submit the jobs with {}.'.format(
-        click.style('mdbenchmark submit', bold=True)))
+    console.info('Finished generating all benchmarks.')
+    console.info(
```
kain88-de: can be one line. If you use an explicit newline `\n`

mdbenchmark/submit.py
```diff
@@ -95,8 +92,9 @@ def submit(directory, force_restart):
         bundles_to_start = bundles_not_yet_started
 
     engine_cmd = get_engine_command()
-    click.echo('Submitting a total of {} benchmarks.'.format(
-        click.style(str(len(bundles_to_start)), bold=True)))
+    console.info(
+        'Submitting a total of {} benchmarks.',
+        placeholders=str(len(bundles_to_start)))
```
kain88-de: the string conversion isn't necessary anymore.

mdbenchmark/utils.py
```diff
@@ -57,7 +59,7 @@ def get_possible_hosts():
 
 def print_possible_hosts():
     all_hosts = get_possible_hosts()
-    click.echo('Available host templates:')
+    console.info('Available host templates:')
     for h in all_hosts:
         click.echo('{}'.format(h))
```
kain88-de: shouldn't this also be `console.info`?

mdbenchmark/console.py
```diff
@@ -0,0 +1,87 @@
...
+                    filehandler=None,
+                    args=None,
+                    bold=True,
+                    fg=None,
+                    bg=None,
+                    underline=None,
+                    blink=None,
+                    **kwargs):
+    """Wrapper to consolidate all click.echo() calls.
+
```
kain88-de: please add a docstring explaining the usage of this wrapper.

mdbenchmark/console.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    if args is None:
+        args = []
+
+    if prefix is not None:
+        message = '{} {}'.format(prefix, message)
+
+        if args:
```
kain88-de: wrong indentation level

mdbenchmark/analyze.py
```diff
@@ -142,20 +143,17 @@ def analyze(directory, plot, ncores):
         df.loc[i] = analyze_run(sim)
 
     if df.isnull().values.any():
-        click.echo(
-            '{}\tWe were not able to gather informations for all systems.\n\t'
+        console.warn(
+            '\tWe were not able to gather informations for all systems.\n\t'
```
kain88-de: why start with a tab here?

mimischi: We are printing several lines to the user here. I wanted everything to be indented after the `WARNING` prefix, so we also need to indent the very first line with a tab.

kain88-de: you can also just print one line. I don't see a reason why this sentence should be broken up on the terminal.

mdbenchmark/analyze.py
```diff
@@ -169,17 +167,13 @@ def analyze(directory, plot, ncores):
...
         uniqueness = df.apply(lambda x: x.nunique())
         if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
-            click.echo(
-                '{} Cannot plot benchmarks for more than one GROMACS module'
-                ' and/or host.'.format(
-                    click.style('ERROR', fg='red', bold=True)))
-            sys.exit(1)
+            console.error(
+                'Cannot plot benchmarks for more than one GROMACS module'
+                ' and/or host.')
```
kain88-de: It is common practice to have trailing whitespaces on multiline strings. This is true across programming languages.

PR #48, opened 2018-02-24
[WIP] Guide user through benchmark generation

mimischi:
If the user calls `mdbenchmark generate` without any options or with some options missing, we will prompt them for these. We currently validate the host template name and tell the user which hosts are available. Further we ask whether the defined modules are correct -- if not, they can redo these questions.

- Prompt user for all required values to generate benchmarks.
- Ask whether the module selection is correct (no validation whether modules are available).
- Check that the host file exists.

PR Checklist
------------
 - [ ] Unit / integration tests

### DIFF
--- a/mdbenchmark/generate.py
+++ b/mdbenchmark/generate.py
@@ -135,7 +135,6 @@ def check_file_available(ctx=None, param=None, value=''):
     '--gpu',
     is_flag=True,
     prompt='Should the benchmarks run on GPUs?',
-    is_eager=True,
     help='Use GPUs for benchmark.',
     show_default=True)
 @click.option(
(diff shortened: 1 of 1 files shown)


mimischi:
We probably need to think more about this. I think we _really_ should implement this feature, as even I forget the syntax for specific options all the time. Getting guided through the process of benchmark generation is nice. Nonetheless we should enable power users to tell the system not to ask them too many questions. In the current implementation the user will always be prompted to define the min- and max-nodes, as well as the run time -- even if you type `mdbenchmark generate --host draco --name md --module gromacs/2016.4`.

Currently I am not sure how we could implement another option, i.e. `--skip` to skip all prompts that have default values.

kain88-de:
> as even I forget the syntax for specific options all the time

I do not know the command line options for 95% of the tools I use on a regular basis. I tend to use the `--help` option a lot and read man pages all the time. It's quite natural to forget the command line switches particularly for tools that aren't used on a daily basis. This also applies to libraries I have written.

mimischi:
Similar to #52, we could think about moving this from `mdbenchmark generate` to some other command (`mdbenchmark wizard` :tophat: :star2:). This way we do not break the API and users who want to just use the usual `mdbenchmark generate` can do so, but new users would be pointed to the new command.

Thoughts? @MSiggel

mimischi:
This is super stale and I do not have time to work on this. If anyone wants to get this going again, feel free!

PR #49, opened 2018-02-24
Add validation to module name

### LINKED ISSUE #47: Generate Function Arguments
Currently we generate individual datreants and folders with 
```python 
def write_bench(top, tmpl, nodes, gpu, module, tpr, name, host, time):
```
the `tpr` argument is very gromacs specific and should be removed in the future once NAMD is implemented. It currently does not impair the code but is unused in the when running this for NAMD.

mimischi:
- Added check of module name to `mdbenchmark.generate`. When given a module name, we try to find it on the system. Users can skip validation with `--skip-validation`. 
- Added `--skip-validation` option to `mdbenchmark.generate`.
- Resolves #47.

PR Checklist
------------
 - [x] Added unit / integration tests.


mdbenchmark/generate.py
```diff
@@ -103,10 +115,40 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
         engine = detect_md_engine(m)
 
+        # Check the current module
+        validated_module = validate_module_name(
+            m, available_modules=available_modules)
+
+        # If we are unable to find the modules path on the system or if the
+        # user told us to skip the validation, we do so.
+        if (validated_module is None) or skip_validation:
+            console.warn('Not performing module name validation.')
```
mimischi: This warning will be printed for every module that we define. We probably should put this outside the for-loop so it is only printed once.

mdbenchmark/generate.py
```diff
@@ -103,10 +115,40 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
...
+        # If the validation fails, we throw an error and exit the script.
+        elif not validated_module:
```
mimischi: We currently do not test this case.

mimischi:
@kain88-de Any thoughts? Besides my two comments, this PR is feature complete.

We could think about moving the code from `mdbenchmark/validate.py` to `mdbenchmark/mdengines/__init__.py` where it would live with `detect_md_engine()`.

mdbenchmark/generate.py
```diff
@@ -103,10 +115,40 @@ def generate(name, gpu, module, host, min_nodes, max_nodes, time, list_hosts):
             'If you use the {} option make sure you use the GPU compatible NAMD module!',
             '--gpu')
 
+    # Grab list of all available and supported modules
+    available_modules = get_available_modules()
```
kain88-de: why not filter the list modules here?

mimischi: Do you mean we should duplicate the `for m in module:` code? Otherwise we cannot do the validation of module names, as we need to iterate through the list.

mimischi:
@kain88-de What is wrong with failing tests on python 2.7? What's wrong about that import statement?

mdbenchmark/generate.py
kain88-de:
```python
modules = [m for m in modules if m in available_modules]
```

Now all modules are actually available. If you want to warn the user for missing ones you can so

```python
requested_modules = set(modules)
modules = [...]
if requested_modules.difference(modules):
    console.warn(" {} are not available modules".format(requested_modules.difference(modules))

```

This if-clause should only be activated if the difference is not empty. If it is empty all modules are available.

mimischi:
This became way bigger than anticipated. I've done a few things in the latest commits:

* Extracted all argument validation for the `generate` CLI command into a new `validate.py`. This way we can test the exceptions separately in `test_validate.py`.
  * Also this cuts down on the size of the `generate` function.
* Added a new module name validation block inside the`generate` function.
  1) It first checks whether the requested MD engine is supported. Throws an error if not.
  2) Then it grabs all available modules and compares them to the requested ones.
  3) It then throws an error for all modules that it could not find and shows all available options to the user.

Several things that we could consider:
1) Refactor the module name validation into some different file. Either`validate.py` or `mdengines/__init__.py` or create another `mdengines/validate.py`.
2) Change the style on how we inform the user. The current messages are pretty repetitive and pretty verbose. This could use some styling.
3) Finally clean up the tests. I really do not like the state of our tests, but I'm too inexperienced with the intricacies of `pytest` to use it to its full extent. I'm sure we can cut down on some of the spaghetti code, e.g. [here](https://github.com/bio-phys/MDBenchmark/blob/validation/mdbenchmark/tests/mdengines/test_init.py#L71-L75) and [here](https://github.com/bio-phys/MDBenchmark/blob/validation/mdbenchmark/tests/mdengines/test_gromacs.py#L115-L121). Lots of test code is repetitive (e.g., calls to `cli_runner` could probably be cut-down with some custom fixture?).

kain88-de:
Sorry about closing this. That was an error

mimischi:
How is one supposed to test the callback functions?

mimischi:
Phew. Refactoring code after the last merge wasn't as exciting as I was hoping for. Everything should be working now.

I still want to make sure that we are testing everything correctly. Quickly went through the CLI manually and found a major bug which I now fixed. When I've finished that, this should finally be ready to merge.

kain88-de:
you can remove the validation tests and clean up the history a bit.

mimischi:
Which validation tests? The contents of `test_validate.py`? I would keep those, because we want to make sure that we throw the expected error messages.

kain88-de:
or move them to the appropriate files. There is no file called `validate.py`

mimischi:
The PR is done. @kain88-de I assume we should keep the merge and just squash the commits in between, right?

mdbenchmark/tests/test_generate.py
```diff
@@ -20,12 +20,28 @@
 import os
 
 import pytest
+from click import exceptions
+
 from mdbenchmark import cli
 from mdbenchmark.ext.click_test import cli_runner
+from mdbenchmark.generate import (validate_hosts, validate_module,
+                                  validate_name, validate_number_of_nodes)
+from mdbenchmark.mdengines import normalize_modules
```
kain88-de: this last function should be tested in `test_mdengines`

kain88-de: It is currently not directly tested in this module either

PR #52, opened 2018-02-27
Update plotting interface

### LINKED ISSUE #24: Extend Plotting options

kain88-de:
Fixes #24

Updated plotting interface. `mdbenchmark analyze --plot` is not deprecated. Use `mdbenchmark plot` instead.

The new workflow for plotting data is as follows:

1) Use `mdbenchmark analyze` to generate a CSV output file.
2) Use `mdbenchmark plot --csv name_of_generated_file.csv` to plot the data.

Consult `mdbenchmark <command> --help` for options to filter your data accordingly.

PR Checklist
------------

### DIFF
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -17,100 +17,15 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with MDBenchmark.  If not, see <http://www.gnu.org/licenses/>.
-import os
-import sys
-from glob import glob
-
 import click
 import mdsynthesis as mds
 import numpy as np
 import pandas as pd
-from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
-from matplotlib.figure import Figure
 
 from . import console
 from .cli import cli
 from .mdengines import detect_md_engine, utils
-from .utils import (calc_slope_intercept, generate_output_name, guess_ncores,
-                    lin_func)
-
-
-def plot_analysis(df, ncores):
-    # We have to use the matplotlib object-oriented interface directly, because
-    # it expects a display to be attached to the system, which we don't on the
-    # clusters.
-    f = Figure()
-    FigureCanvas(f)
-    ax = f.add_subplot(111)
-
-    # Remove NaN values. These are missing ncores/performance data.
-    df = df.dropna()
-
-    max_x = df['nodes'].max()
-    max_y = df['ns/day'].max()
-
-    if not max_y:
-        max_y = 50
-
-    y_tick_steps = 10
-    if max_y > 100:
-        y_tick_steps = 20
-
-    x = np.arange(1, max_x + 1, 1)
-
-    if not df[~df['gpu']].empty:
-        cpu_data = df[(~df['gpu'])].sort_values('nodes').reset_index()
-        ax.plot(cpu_data['ns/day'], '.-', ms='10', color='C0', label='CPU')
-        slope, intercept = calc_slope_intercept(x[0], cpu_data['ns/day'][0],
-                                                x[1], cpu_data['ns/day'][1])
-        ax.plot(
-            x - 1,
-            lin_func(x, slope, intercept),
-            ls='dashed',
-            color='C0',
-            alpha=0.5)
-
-    gpu_data = df[(df['gpu'])].sort_values('nodes').reset_index()
-    if not gpu_data.empty:
-
-        ax.plot(gpu_data['ns/day'], '.-', ms='10', color='C1', label='GPU')
-        slope, intercept = calc_slope_intercept(x[0], gpu_data['ns/day'][0],
-                                                x[1], gpu_data['ns/day'][1])
-        ax.plot(
-            x - 1,
-            lin_func(x, slope, intercept),
-            ls='dashed',
-            color='C1',
-            alpha=0.5)
-
-    ax.set_xticks(x - 1)
-    ax.set_xticklabels(x)
-
-    axTicks = ax.get_xticks()
-
-    ax2 = ax.twiny()
-    ax2.set_xticks(axTicks)
-    ax2.set_xbound(ax.get_xbound())
-    if ncores is not None:
-        console.info(
-            "Ncores overwritten from CLI. Ignoring values from simulation logs for plot."
-        )
-        ax2.set_xticklabels(x for x in (axTicks + 1) * ncores)
-    else:
-        ax2.set_xticklabels(cpu_data['ncores'])
-
-    ax.set_xlabel('Number of nodes')
-    ax.set_ylabel('Performance [ns/day]')
-
-    ax.set_yticks(np.arange(0, (max_y + (max_y * 0.5)), y_tick_steps))
-    ax.set_ylim(ymin=0, ymax=(max_y + (max_y * 0.2)))
-
-    ax2.set_xlabel('{}'.format('{}\n\nCores'.format(df['host'][0])))
-
-    ax.legend()
-
-    f.tight_layout()
-    f.savefig('runtimes.pdf', format='pdf')
+from .utils import generate_output_name
 
 
 @cli.command()
@@ -182,19 +97,20 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
-        df = pd.read_csv(output_name)
+        raise NotImplemented
+        # df = pd.read_csv(output_name)
 
-        # We only support plotting of benchmark systems from equal hosts /
-        # with equal settings
-        uniqueness = df.apply(lambda x: x.nunique())
-        if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
-            console.error(
-                'Cannot plot benchmarks for more than one GROMACS module '
-                'and/or host.')
+        # # We only support plotting of benchmark systems from equal hosts /
+        # # with equal settings
+        # uniqueness = df.apply(lambda x: x.nunique())
+        # if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
+        #     console.error(
+        #         'Cannot plot benchmarks for more than one GROMACS module '
+        #         'and/or host.')
 
-        # Fail if we have no values at all. This should be some edge case when
-        # a user fumbles around with the datreant categories
-        if df['gpu'].empty and df[~df['gpu']].empty:
-            console.error('There is no data to plot.')
+        # # Fail if we have no values at all. This should be some edge case when
+        # # a user fumbles around with the datreant categories
+        # if df['gpu'].empty and df[~df['gpu']].empty:
+        #     console.error('There is no data to plot.')
 
-        plot_analysis(df, ncores)
+        # plot_analysis(df, ncores)
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -18,13 +18,17 @@
 # You should have received a copy of the GNU General Public License
 # along with MDBenchmark.  If not, see <http://www.gnu.org/licenses/>.
 import click
+import matplotlib.pyplot as plt
 import mdsynthesis as mds
 import numpy as np
 import pandas as pd
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
 
 from . import console
 from .cli import cli
 from .mdengines import detect_md_engine, utils
+from .plot import plot_line
 from .utils import generate_output_name
 
 
@@ -97,20 +101,27 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
-        raise NotImplemented
-        # df = pd.read_csv(output_name)
+        console.warn(
+            'This feature is now outdated.'
+            'Please use mdbenchmark plot to access all plotting features'
+            'Future versions will not support this feature anymore.')
 
-        # # We only support plotting of benchmark systems from equal hosts /
-        # # with equal settings
-        # uniqueness = df.apply(lambda x: x.nunique())
-        # if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
-        #     console.error(
-        #         'Cannot plot benchmarks for more than one GROMACS module '
-        #         'and/or host.')
+        fig = Figure()
+        FigureCanvas(fig)
+        ax = fig.add_subplot(111)
 
-        # # Fail if we have no values at all. This should be some edge case when
-        # # a user fumbles around with the datreant categories
-        # if df['gpu'].empty and df[~df['gpu']].empty:
-        #     console.error('There is no data to plot.')
+        df = pd.read_csv(output_name)
 
-        # plot_analysis(df, ncores)
+        df_sel = 'nodes'
+        gb = df.groupby(['gpu', 'module', 'host'])
+        groupby = ['gpu', 'module', 'host']
+        for key, df in gb:
+            label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
+            plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
+
+        ax.set_xlabel('Number of Nodes')
+        ax.set_ylabel('Performance [ns/day]')
+        ax.legend()
+
+        fig.tight_layout()
+        fig.savefig('runtimes.pdf', format='pdf')
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -54,11 +54,8 @@ plt.switch_backend('agg')
     'benchmarks log file.',
     show_default=True)
 @click.option(
-    '-o',
-    '--output-name',
-    default=None,
-    help="Name of the output .csv file.",
-    type=str)
+    "-o", "--output-name", default=None, help="Name of the output CSV file.", type=str
+)
 def analyze(directory, plot, ncores, output_name):
     """Analyze finished benchmarks."""
     bundle = mds.discover(directory)
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -103,7 +103,8 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
-        console.warn('DEPECATED Please use \'{}\' in the future', 'mdbenchmark plot')
+        console.warn('This feature is deprecated. Please use \'{}\' in the future.',
+                     'mdbenchmark plot')
 
         fig = Figure()
         FigureCanvas(fig)
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -28,7 +28,7 @@ from matplotlib.figure import Figure
 from . import console
 from .cli import cli
 from .mdengines import detect_md_engine, utils
-from .plot import plot_line
+from .plot import plot_over_group
 from .utils import generate_output_name
 
 
@@ -101,27 +101,15 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
-        console.warn(
-            'This feature is now outdated.'
-            'Please use mdbenchmark plot to access all plotting features'
-            'Future versions will not support this feature anymore.')
+        console.warn('DEPECATED Please use \'{}\' in the future', 'mdbenchmark plot')
 
         fig = Figure()
         FigureCanvas(fig)
         ax = fig.add_subplot(111)
 
         df = pd.read_csv(output_name)
-
-        df_sel = 'nodes'
-        gb = df.groupby(['gpu', 'module', 'host'])
-        groupby = ['gpu', 'module', 'host']
-        for key, df in gb:
-            label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
-            plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
-
-        ax.set_xlabel('Number of Nodes')
-        ax.set_ylabel('Performance [ns/day]')
-        ax.legend()
+        ax = plot_over_group(df, plot_cores=False, ax=ax)
+        lgd = ax.legend(loc='upper center', bbox_to_anchor=(0.5, -0.175))
 
         fig.tight_layout()
-        fig.savefig('runtimes.pdf', format='pdf')
+        fig.savefig('runtimes.pdf', type='pdf', bbox_extra_artists=(lgd,), bbox_inches='tight')
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -33,6 +33,7 @@ from .utils import generate_output_name
 
 plt.switch_backend('agg')
 
+
 @cli.command()
 @click.option(
     '-d',
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -31,6 +31,7 @@ from .mdengines import detect_md_engine, utils
 from .plot import plot_over_group
 from .utils import generate_output_name
 
+plt.switch_backend('agg')
 
 @cli.command()
 @click.option(
--- a/mdbenchmark/plot.py
+++ b/mdbenchmark/plot.py
@@ -1,0 +1,94 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4 fileencoding=utf-8
+#
+# MDBenchmark
+# Copyright (c) 2017 Max Linke & Michael Gecht and contributors
+# (see the file AUTHORS for the full list of names)
+#
+# MDBenchmark is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# MDBenchmark is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MDBenchmark.  If not, see <http://www.gnu.org/licenses/>.
+import pandas as pd
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
+import click
+
+from .cli import cli
+from .utils import calc_slope_intercept, lin_func
+
+
+def plot_line(df, label, ax=None):
+    if ax is None:
+        ax = plt.gca()
+
+    p = ax.plot('nodes', 'ns/day', '.-', data=df, ms='10', label=label)
+    color = p[0].get_color()
+    slope, intercept = calc_slope_intercept(
+        (df['nodes'].iloc[0], df['ns/day'].iloc[0]), (df['nodes'].iloc[1],
+                                                      df['ns/day'].iloc[1]))
+    # avoid a label and use values instead of pd.Series
+    ax.plot(
+        df['nodes'],
+        lin_func(df['nodes'].values, slope, intercept),
+        ls='--',
+        color=color,
+        alpha=.5)
+    return ax
+
+
+def plot_over_group(df, groupby, ax=None):
+    # plot all lines
+    gb = df.groupby(groupby)
+    for key, df in gb:
+        label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
+        plot_line(ax=ax, df=df, label=label)
+
+    # style axes
+    ax.set_xlabel('Number of nodes')
+    ax.set_ylabel('Performance [ns/day]')
+    ax.legend()
+
+    ax2 = ax.twiny()
+    ax1_xticks = ax.get_xticks()
+    ax2.set_xticks(ax1_xticks)
+    ax2.set_xbound(ax.get_xbound())
+    ax2.set_xticklabels(gb.get_group(list(gb.groups.keys())[0])['ncores'])
+    ax2.set_xlabel('{}'.format('{}\n\nCores'.format(df['host'][0])))
+
+    return ax
+
+
+@cli.command()
+@click.option('--csv', help='name of csv file')
+@click.option(
+    '--groupby',
+    '-g',
+    type=str,
+    help='columns to groupby for plot as comma separated string',
+    default='gpu, module',
+    show_default=True)
+def plot(csv, groupby):
+    """Plot nice things"""
+    df = pd.read_csv(csv, index_col=0)
+    # Remove NaN values. These are missing ncores/performance data.
+    df = df.dropna()
+    # We have to use the matplotlib object-oriented interface directly, because
+    # it expects a display to be attached to the system, which we don't on the
+    # clusters.
+    fig = Figure()
+    FigureCanvas(fig)
+    ax = fig.add_subplot(111)
+    # remove whitespace
+    groupby = [s.strip() for s in groupby.split(',')]
+    plot_over_group(df, groupby, ax=ax)
+    fig.savefig('results.pdf')
--- a/mdbenchmark/plot.py
+++ b/mdbenchmark/plot.py
@@ -83,6 +83,9 @@ def filter_dataframe_for_plotting(df, host_name, module_name, gpu, cpu):
     elif not cpu and not gpu:
         console.error("CPU and GPU not set. Nothing to plot. Exiting.")
 
+    if df.empty:
+        console.error("Your filtering led to an empty dataset. Exiting.")
+
     df_filtered_hosts = df[df['host'].isin(host_name)]
     df_unique_hosts = np.unique(df_filtered_hosts['host'])
 
--- a/mdbenchmark/plot.py
+++ b/mdbenchmark/plot.py
@@ -94,19 +94,21 @@ def filter_dataframe_for_plotting(df, host_name, module_name, gpu, cpu):
             ', '.join(np.unique(df['host']))))
 
     if not host_name:
-        console.info("Plotting all hosts in csv.")
+        console.info("Plotting all hosts in input file.")
     else:
         df = df_filtered_hosts
         console.info("Data for the following hosts will be plotted: {}".format(
             ', '.join(df_unique_hosts)))
 
     for module in module_name:
-        if module in ['gromacs', 'namd']:
-            console.info("Plotting all modules for engine {}.", module)
-        elif module in df['module'].tolist():
-            console.info("Plotting module {}.", module)
-        elif module not in df['module'].tolist():
-            console.error("The module {} does not exist in your data. Exiting.", module)
+        if module in ["gromacs", "namd"]:
+            console.info("Plotting all modules for engine '{}'.", module)
+        elif module in df["module"].tolist():
+            console.info("Plotting module '{}'.", module)
+        elif module not in df["module"].tolist():
+            console.error(
+                "The module '{}' does not exist in your data. Exiting.", module
+            )
 
     if not module_name:
         console.info("Plotting all modules in your input data.")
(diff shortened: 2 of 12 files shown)


mdbenchmark/plot.py
```diff
@@ -0,0 +1,94 @@
...
+@cli.command()
+@click.option('--csv', help='name of csv file')
+@click.option(
+    '--groupby',
+    '-g',
+    type=str,
+    help='columns to groupby for plot as comma separated string',
+    default='gpu, module',
+    show_default=True)
+def plot(csv, groupby):
```
kain88-de: need to add an option for output name.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,94 @@
...
+    ax.set_xlabel('Number of nodes')
+    ax.set_ylabel('Performance [ns/day]')
+    ax.legend()
+
+    ax2 = ax.twiny()
+    ax1_xticks = ax.get_xticks()
+    ax2.set_xticks(ax1_xticks)
+    ax2.set_xbound(ax.get_xbound())
+    ax2.set_xticklabels(gb.get_group(list(gb.groups.keys())[0])['ncores'])
+    ax2.set_xlabel('{}'.format('{}\n\nCores'.format(df['host'][0])))
```
kain88-de: This axes labeling here can get very tricky very quickly. For example, if we allow for more then one host in the plot the number of cores per node can be different. In such a case we would only want to plot one xlabel with the number of cores for comparison. We can't do our current trick anymore of also plotting the number of nodes.

kain88-de: Also the different benchmarks that are in the csv file could have been run on a different range of nodes. The plot_line function already takes this into account.

mdbenchmark/plot.py
kain88-de: also an option to exclude some rows would be nice to have.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,94 @@
...
+from matplotlib.figure import Figure
+import click
+
+from .cli import cli
+from .utils import calc_slope_intercept, lin_func
+
+
+def plot_line(df, label, ax=None):
+    if ax is None:
+        ax = plt.gca()
```
mimischi: What is the advantage of `plt.gca()` over the usual `plt.subplots()`?

kain88-de: it grabs the last used axes objects *or* creates a new figure. It's a common idiom in matplotlib libraries.

mimischi:
@kain88-de What is the status of this PR? Are you going to finish it or is it just a proof-of-concept?

MSiggel:
The user can now provide multiple csv files and plot what he needs from these. The options are now given as individual flags in close analogy to generate. The initial df is filtered and then the remaining data plotted into one file.

MSiggel:
I will include Engines by simply giving only the engine name as module name e.g gromacs instead of gromacs/2018.

mdbenchmark/analyze.py
```diff
@@ -177,19 +93,20 @@ def analyze(directory, plot, ncores, output_name):
...
-                'Cannot plot benchmarks for more than one GROMACS module '
-                'and/or host.')
-
-        # Fail if we have no values at all. This should be some edge case when
-        # a user fumbles around with the datreant categories
-        if df['gpu'].empty and df[~df['gpu']].empty:
-            console.error('There is no data to plot.')
-
-        plot_analysis(df, ncores)
+        raise NotImplemented
```
mimischi: Instead of this, we should use `console.error` to notify the user to use `mdbenchmark plot` instead of `mdbenchmark analyze --plot`.

mimischi: Also the commented block below can be deleted

mdbenchmark/plot.py
```diff
@@ -0,0 +1,210 @@
...
+    multiple=True,
+    show_default=True)
+@click.option(
+    '--output-name',
+    '-o',
+    default=None,
+    help="name of output files",
+    show_default=True)
+@click.option(
+    '--output-type',
```
mimischi: Can we use `multiple` here? This way the user could generate a plot with different file types in one go.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,210 @@
...
+    show_default=True,
+    default=False)
+def plot(csv, output_name, output_type, host_name, module_name, gpu, cpu, plot_nodes):
+    """Plot nice things"""
+    print(plot_nodes)
+    df_list = []
+    for c in csv:
+        tmp_df = pd.read_csv(c, index_col=0)
+        # append df_list
+        df_list.append(tmp_df)
```
mimischi: We should be able to do this instead:

```python
df = pd.DataFrame()

for c in csv:
    tmp_df = pd.read_csv(c, index_col=0)
    df.append(tmp_df, ignore_index=True)
```

Then we do not need to do `df = pd.concat(df_list)` on line 146.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,210 @@
...
+    df = df.dropna()
+
+    # preprocess the commandline entries
+
+    df_module_list = df['module'].tolist()
+    processed_module_names = []
+    print(df_module_list)
+    for module in module_name:
+        if module in ['namd', 'gromacs']:
+            real_module_names = [s for s in df_module_list if module in s]
```
mimischi: What exactly are you trying to check here?

mdbenchmark/plot.py
```diff
@@ -0,0 +1,210 @@
...
+        elif module in df_module_list:
+            processed_module_names.append(module)
+        elif module not in df_module_list:
+            console.warn("The module {} does not exist in your data. Exiting",
```
mimischi: We have `console.error`, when throwing messages that exit the program.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,210 @@
...
+    for host in host_name:
+        if host not in host_list:
+            console.warn("The host {} does not exist in your csv data. Exiting.",
+                         host)
+
+    gpu_list = []
+    if gpu is True:
+        gpu_list.append(True)
+    if cpu is True:
+        gpu_list.append(False)
```
mimischi: Should this be `cpu_list`?

mdbenchmark/plot.py
MSiggel: this is the check which is envoked when the user only gives an engine name. Then all modules which contain this engine name are used for the module selection e.g gromacs/5.1.4 gromacs/2016 would both be selected

mdbenchmark/plot.py
MSiggel: we can rename this to cpu_gpu or something. since for gpu or cpu either false or true is listed in the dataframe i make a list containing either one of them or both if the user wants to plot both gpu and cpu

mdbenchmark/plot.py
MSiggel: i would like to give the option to only give "gromacs" as option to get a plot with all gromacs version which are in a df. One would need some kind of string match to do this

mdbenchmark/plot.py
MSiggel: I tried this and one can't append an empty df.

mdbenchmark/plot.py
```diff
@@ -47,24 +47,21 @@ def plot_line(df, df_sel, label, ax=None):
...
 
 
-def plot_over_group(df, plot_nodes, ax=None):
+def plot_over_group(df, plot_cores, ax=None):
     # plot all lines
-    if plot_nodes is False:
-        df_sel = "nodes"
-    else:
-        df_sel = "ncores"
+    df_sel = 'ncores' if 'plot_cores' else 'nodes'
```
kain88-de: Did you test this?

MSiggel:
we need to think about testing the plot function. I am not really sure how to do this because the main function only returns the pdf,png file. 
I would like to be able to test different combinations of selections with parametrize to make sure the filters work properly but I need to be able to compare the plotted df with some reference df.
Do we need to put the df processing in a new file?
Or we compare the output picture bit by bit...
@mimischi thoughts on this?

mdbenchmark/plot.py
```diff
@@ -0,0 +1,213 @@
...
+        gpu_cpu_list.append(False)
+
+    # here I split all data frames into the posible smallest segments
+    # this is necessary so we can plot all individually
+    split_df = df.groupby(['gpu', 'module', 'host'])
+
+    print(split_df)
+    # here I initialize the list which will be plotted
+    df_list = []
+    for key, df in split_df:
```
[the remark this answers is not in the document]
MSiggel: I'm not sure I know what you mean exactly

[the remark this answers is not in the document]
kain88-de: this is some additional filtering on the dataframe. It seems cumbersome that you have two loops with if-else clauses. There should be an easier way to do this with pandas.

mdbenchmark/plot.py
kain88-de: I get that this would be the outcome. What is the possible use of this? How often would you want to use this?
Remember that you will need to support this for a long time going forward.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,246 @@
...
+    fig = Figure()
+    FigureCanvas(fig)
+    ax = fig.add_subplot(111)
+    plot_over_group(df, plot_cores, ax=ax)
+
+    if output_name is None:
+        output_name = generate_output_name(output_type)
+    if output_type not in output_name:
+        output_name = '{}.{}'.format(output_name, output_type)
+    fig.savefig(output_name)
```
mimischi: Although `savefig` determines the filetype from the given file name, I would also explicitly pass it: `fig.savefig(output_name, type=output_type)`

mdbenchmark/analyze.py
```diff
@@ -177,19 +98,26 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
+        console.warn(
```
mimischi: Are we using `console.warn` because we assume that users expect `mdbenchmark analyze --plot` to do something? I'd rather use `console.error` and exit, but can relate to the backwards compatibility.

Maybe we should announce a planned deprecation? I do not think keeping the feature for a longer time is that great.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def plot_over_group(df, plot_cores, ax=None):
+    # plot all lines
+    df_sel = 'ncores' if plot_cores else 'nodes'
+
+    gb = df.groupby(['gpu', 'module', 'host'])
+    groupby = ['gpu', 'module', 'host']
+    for key, df in gb:
+        label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
```
mimischi: We need to think about the label text. The current approach isn't great.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,246 @@
...
+        plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
+
+    # style axes
+    if plot_cores:
```
mimischi: We could shrink the `if-else` to:

```python
xlabel = 'cores' if plot_cores else 'nodes'
ax.set_xlabel('Number of {}'.format(xlabel))
```

mdbenchmark/plot.py
```diff
@@ -0,0 +1,246 @@
...
+        ax.set_xlabel('Number of Cores')
+    else:
+        ax.set_xlabel('Number of Nodes')
+
+    ax.set_ylabel('Performance [ns/day]')
+    ax.legend()
+
+    # ax2 = ax.twiny()
```
mimischi: The commented code needs to go before we merge.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,246 @@
...
+@cli.command()
+@click.option(
+    '--csv',
+    help='name of csv file',
+    multiple=True,
+    show_default=True)
+@click.option(
+    '--output-name',
+    '-o',
+    default=None,
```
mimischi: Isn't `None` the default for the `default` argument in `click`?

MSiggel:
As discussed, filtering the data frame is now done by using pandas commands and the previously used crude workarounds were removed.
The error and info message handling still needs improvement and the tests for plot need to be written and updated respectively.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+    if not gpu:
+        df = df[df['gpu'].isin([False])]
+
+    if len(df[df['host'].isin(host_name)]) > 0:
+        df = df[df['host'].isin(host_name)]
+        console.info('Data for the following hosts will be plotted: {}.', set(host_name))
+    elif not host_name:
+            # dataframe stays the same and all hosts will be plotted
+            console.info('All hosts will be plotted.')
+    elif len(df[df['host'].isin(host_name)]) == 0:
```
kain88-de: `if host_name not in df['host']` is even simpler and works two. The same goes for the initial check.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+
+    df = pd.concat(df_list)
+
+    # Remove NaN values. These are missing ncores/performance data.
+    df = df.dropna()
+
+    if not gpu:
+        df = df[df['gpu'].isin([False])]
+
+    if len(df[df['host'].isin(host_name)]) > 0:
```
kain88-de: more suited if the argument to `isin` is iterable. A more intuitive solution is `if host_name in df.host`.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+            # dataframe stays the same and all hosts will be plotted
+            console.info('All hosts will be plotted.')
+    elif len(df[df['host'].isin(host_name)]) == 0:
+        # filter host names that don't exist
+        console.error('The host {} does not exist in your csv data. Exiting.', host)
+
+    for module in module_name:
+        if module in ['gromacs', 'namd']:
+            console.info("plotting all modules for engine {}.", module)
+        elif len(df[df['module'].isin([module])]) == 0:
```
kain88-de: see above

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+        # append df_list
+        df_list.append(tmp_df)
+
+    df = pd.concat(df_list)
+
+    # Remove NaN values. These are missing ncores/performance data.
+    df = df.dropna()
+
+    if not gpu:
+        df = df[df['gpu'].isin([False])]
```
kain88-de: `df = df[not df.gpu]`

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+    default=True)
+@click.option(
+    '--plot-cores',
+    help="Plot performance per core instead of nodes",
+    show_default=True,
+    is_flag=True)
+def plot(csv, output_name, output_type, host_name, module_name, gpu, cpu, plot_cores):
+    """Generate plots from csv files"""
+
+    df = pd.DataFrame()
```
kain88-de: not needed.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,271 @@
...
+    if len(csv) == 0:
+        console.error("No csv file specified. use --csv to specify a file.")
+
+    for c in csv:
+
+        tmp_df = pd.read_csv(c, index_col=0)
+        # append df_list
+        df_list.append(tmp_df)
+
+    df = pd.concat(df_list)
```
kain88-de: `df = pd.concat([pd.read_csv(c, index_col=0) for c in csv]).dropna()`

mdbenchmark/plot.py
mimischi: I think the official boolean indexing would be `df = df[~df.gpu]`, see `pandas` documentation: https://pandas.pydata.org/pandas-docs/stable/indexing.html#boolean-indexing

mdbenchmark/plot.py
MSiggel: i had to convert the entries to lists to be able to do this

mdbenchmark/analyze.py
```diff
@@ -177,19 +98,26 @@ def analyze(directory, plot, ncores, output_name):
...
         df = pd.read_csv(output_name)
 
-        # We only support plotting of benchmark systems from equal hosts /
-        # with equal settings
-        uniqueness = df.apply(lambda x: x.nunique())
-        if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
-            console.error(
-                'Cannot plot benchmarks for more than one GROMACS module '
-                'and/or host.')
+        df_sel = 'nodes'
```
mimischi: What do we need this for? Any reason to set the variable here, instead of in the `plot_line` call further below?

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+@click.option(
+    '--output-name',
+    '-o',
+    default=None,
+    help="name of output files",
+    show_default=True)
+@click.option(
+    '--output-type',
+    '-t',
+    help="file extension for plot outputs",
```
mimischi: All help texts should be put into the same style as the rest of the package. At the very last capitalize the first letter of a sentence.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    help="hostname",
+    show_default=True)
+@click.option(
+    '--module-name',
+    '-h',
+    default=None,
+    multiple=True,
+    help="module name or engine name (gromacs, namd)",
+    show_default=True)
+@click.option(
```
mimischi: This is a duplicate of the option on line 88-94.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    help="file extension for plot outputs",
+    type=click.Choice(['png', 'pdf', 'svg', 'ps']),
+    show_default=True,
+    default='png')
+@click.option(
+    '--host-name',
+    '-h',
+    default=None,
+    multiple=True,
+    help="hostname",
```
mimischi: Same as above.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    show_default=True)
+@click.option(
+    '--module-name',
+    '-h',
+    default=None,
+    multiple=True,
+    help="module name or engine name (gromacs, namd)",
```
mimischi: Same as above.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+@click.option(
+    '--host-name',
+    '-h',
+    default=None,
+    multiple=True,
+    help="hostname",
+    show_default=True)
+@click.option(
+    '--gpu/--no-gpu',
+    help="plot data for GPU runs",
```
mimischi: Same as above.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    show_default=True,
+    default=False)
+@click.option(
+    '--cpu/--no-cpu',
+    help="plot data for CPU runs",
```
mimischi: Same as above.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    ax.set_ylabel('Performance [ns/day]')
+    ax.legend()
+
+    return ax
+
+
+@cli.command()
+@click.option(
+    '--csv',
+    help='name of csv file',
```
mimischi: Aaah.. same as below :sunglasses:

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    default=True)
+@click.option(
+    '--plot-cores',
+    help="Plot performance per core instead of nodes",
+    show_default=True,
+    is_flag=True)
+def plot(csv, output_name, output_type, host_name, module_name, gpu, cpu, plot_cores):
+    """Generate plots from csv files"""
+
+    if len(csv) == 0:
```
mimischi: `if not csv` should be sufficient. Also you probably should use the `callback` argument for `@click.option`, like we do for `generate` now: https://github.com/bio-phys/MDBenchmark/blob/master/mdbenchmark/generate.py#L99-L103

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+        # this is incase the user decides to set both flags to true
+        console.info("plotting GPU and CPU data.")
+    elif gpu:
+        df = df[df.gpu]
+        console.info("plotting GPU data only.")
+    elif cpu:
+        df = df[~df.gpu]
+        console.info("plotting GPU data only.")
+    else:
+        console.info("plotting GPU and CPU data.")
```
mimischi: Does this case ever trigger? Also we should validate, if `cpu` and `gpu` are both false. In this case, we throw an error. Similar to what is implemented here: https://github.com/bio-phys/MDBenchmark/pull/69/files#diff-5ecad2076077c8bab2ad9de9fb25c45aR51

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+        if module in ['gromacs', 'namd']:
+            console.info("plotting all modules for engine {}.", module)
+        elif module not in df['module'].tolist():
+            console.error("The module {} does not exist in your data. Exiting.", module)
+
+    if not module_name:
+        console.info("plotting all modules in csv.")
+    # this should work but we need to check before whether any of the entered
+    # names are faulty/don't exist
+    if module_name:
```
mimischi: You could piggy back the `if`-clause above an just convert that this one to `else:`

mdbenchmark/plot.py
```diff
@@ -0,0 +1,188 @@
...
+        df = df[df['module'].str.contains('|'.join(module_name))]
+
+    if len(df) == 0:
```
mimischi: Use `if df.empty:`. See: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.empty.html

mdbenchmark/analyze.py
MSiggel: yes, currently this is still usable and I think it makes sense to keep it this way for now until the new version has been properly tested.

mdbenchmark/plot.py
MSiggel: agreed, but I wouldn't leave out any of the information if possible. Maybe it would be possible to move the labels into a box below the plot in form of a little table or something? Or maybe one could even print the dataframe below as reference?

kain88-de: Have a look at the seaborn.FacetGrid code. It can place the legends outside the plotting area as standard.

kain88-de: https://github.com/mwaskom/seaborn/blob/67126155553f8bc014fc178ae9a812f604efaa25/seaborn/axisgrid.py#L83

mdbenchmark/analyze.py
```diff
@@ -177,19 +98,27 @@ def analyze(directory, plot, ncores, output_name):
...
+        gb = df.groupby(['gpu', 'module', 'host'])
+        groupby = ['gpu', 'module', 'host']
```
mimischi: Stylistic change:

```python
groupby = ['gpu', 'module', 'host']
groups = df.groupby(groupby)
```

mdbenchmark/plot.py
mimischi: Yes, we should probably use something like that. I recently had the same problem with one of my plots and stumbled upon this excellent answer on StackOverflow reviewing all possibilities with `matplotlib`: https://stackoverflow.com/a/43439132/4791226

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+    '-t',
+    help="File extension for output file.",
+    type=click.Choice(['png', 'pdf', 'svg', 'ps']),
+    show_default=True,
+    default='png')
+@click.option(
+    '--module-name',
+    '-h',
+    multiple=True,
+    help="Module name or engine name (gromacs, namd) which is plotted. ",
```
mimischi: Minor: there is a trailing white space after the period.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+    show_default=True)
+@click.option(
+    '--host-name',
+    '-h',
+    multiple=True,
+    help="Host name which is plotted.",
+    show_default=True)
```
mimischi: I don't think `show_default` is needed, because there is no default value.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+    default=True)
+@click.option(
+    '--cpu/--no-cpu',
+    help="Plot data for CPU runs.",
+    show_default=True,
+    default=True)
+@click.option(
+    '--plot-cores',
+    help="Plot performance per core instead of nodes",
+    show_default=True,
```
mimischi: Same here.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+@click.option(
+    '--csv',
+    help='Name of csv file to plot.',
+    multiple=True,
+    show_default=True)
+@click.option(
+    '--output-name',
+    '-o',
+    help="Name of output file.",
+    show_default=True)
```
mimischi: Same here.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+
+    return ax
+
+
+@cli.command()
+@click.option(
+    '--csv',
+    help='Name of csv file to plot.',
+    multiple=True,
+    show_default=True)
```
mimischi: Same here.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,178 @@
...
+@click.option(
+    '--plot-cores',
+    help="Plot performance per core instead of nodes",
+    show_default=True,
+    is_flag=True)
+def plot(csv, output_name, output_type, host_name, module_name, gpu, cpu, plot_cores):
+    """Generate plots from csv files"""
+
+    if not csv:
+        console.error("No csv file specified. use --csv to specify a file.")
```
mimischi: Use `click.BadParameter` for this and similar cases. See an example here: https://github.com/bio-phys/MDBenchmark/pull/69/files#diff-5ecad2076077c8bab2ad9de9fb25c45aR56

MSiggel:
I moved the plot ledgend outside of the graph. We can still discuss how to best improve this formatting. 

I also added some tests for the df filtering. I would appreciate some help concerning this test which currently fails:
```
test_plot_filter_dataframe_for_plotting_gpu_and_cpu_fail([...])
```
How do we best catch the error which is generated by click because nothing is plotted and evaluated is such that the test passes?

mdbenchmark/plot.py
```diff
@@ -0,0 +1,185 @@
...
+        label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
+        plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
+
+    # style axes
+    xlabel = 'cores' if plot_cores else 'nodes'
+    ax.set_xlabel('Number of {}'.format(xlabel))
+    ax.set_ylabel('Performance [ns/day]')
+
+    lgd = ax.legend(loc='upper center', bbox_to_anchor=(0.5, -0.175))
+    plt.tight_layout()
```
kain88-de: should be called on the figure outside of this function

mdbenchmark/analyze.py
```diff
@@ -182,19 +103,27 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
+        console.warn(
+            'This feature is now outdated.'
+            'Please use mdbenchmark plot to access all plotting features'
```
kain88-de:
```
'Please use `mdbenchmark plot` to ...
```

mdbenchmark/analyze.py
```diff
@@ -182,19 +103,27 @@ def analyze(directory, plot, ncores, output_name):
...
-        if uniqueness['gromacs'] > 1 or uniqueness['host'] > 1:
-            console.error(
-                'Cannot plot benchmarks for more than one GROMACS module '
-                'and/or host.')
+        df_sel = 'nodes'
+        gb = df.groupby(['gpu', 'module', 'host'])
+        groupby = ['gpu', 'module', 'host']
+        for key, df in gb:
+            label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
+            plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
```
kain88-de: why not `plot.plot_over_group(df, ax=ax)`?

mdbenchmark/analyze.py
```diff
@@ -182,19 +103,27 @@ def analyze(directory, plot, ncores, output_name):
...
 
-        # Fail if we have no values at all. This should be some edge case when
-        # a user fumbles around with the datreant categories
-        if df['gpu'].empty and df[~df['gpu']].empty:
-            console.error('There is no data to plot.')
+        ax.set_xlabel('Number of Nodes')
+        ax.set_ylabel('Performance [ns/day]')
```
kain88-de: ax.set(xlabel=' ... ', ylabel='...')

mdbenchmark/plot.py
```diff
@@ -0,0 +1,185 @@
...
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
+import click
+
+from .cli import cli
+from .utils import calc_slope_intercept, lin_func, generate_output_name
+from . import console
+
+
+def plot_line(df, df_sel, label, ax=None):
```
kain88-de: instead of `df_sel` which doesn't say much about the dataframe selection this is doing, call the variable `x` or `xaxis_col`?

mdbenchmark/plot.py
```diff
@@ -0,0 +1,185 @@
...
+    gb = df.groupby(['gpu', 'module', 'host'])
+    groupby = ['gpu', 'module', 'host']
+    for key, df in gb:
+        label = ' '.join(['{}={}'.format(n, v) for n, v in zip(groupby, key)])
+        plot_line(df=df, df_sel=df_sel, ax=ax, label=label)
+
+    # style axes
+    xlabel = 'cores' if plot_cores else 'nodes'
+    ax.set_xlabel('Number of {}'.format(xlabel))
+    ax.set_ylabel('Performance [ns/day]')
```
kain88-de: you can use the  `ax.set(xlabel=, ylabel=) call.

mdbenchmark/analyze.py
kain88-de: why not turn this into a console info and keep it. As long as this internally uses `plot_over_group` there isn't much overhead.

mdbenchmark/tests/test_plot.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+@pytest.mark.parametrize("host_name", [
+    ('phys'),
+    ('draco')
+])
+def test_plot_filter_dataframe_for_plotting_host_name(cli_runner, tmpdir, data, host_name):
+    """ check whether output types are constructed correctly
+    """
+    with tmpdir.as_cwd():
```
kain88-de: why are you running this in a temporary directory? are you writing files?

mdbenchmark/tests/test_plot.py
```diff
@@ -0,0 +1,234 @@
...
+
+        expected_df = pd.read_csv(data['testcsv.csv'])
+        expected_df = expected_df[expected_df['host'].str.contains(host_name)]
+
+        input_df = pd.read_csv(data['testcsv.csv'])
+
+        # I have to convert the variable to a tuple, otherwise this appears not to work
+        host = (host_name,)
```
kain88-de: yes this doesn't work because the for-loop you have written for hostname validation expects an array-like type. You can also iterate over a string but that will give you the letters of the string.

mdbenchmark/tests/test_plot.py
```diff
@@ -0,0 +1,234 @@
...
+
+        real_df = plot.filter_dataframe_for_plotting(df=input_df, host_name=(), module_name=(), gpu=gpu, cpu=cpu)
+
+        # here we compare the dataframes for any differences
+        assert_frame_equal(input_df, real_df)
+
+
+@pytest.mark.parametrize("gpu,cpu", [
+    (False, False)
+])
```
kain88-de: with only one value there is no need for a parametrize decorator

mdbenchmark/utils.py
```diff
@@ -121,6 +123,6 @@ def guess_ncores():
 def generate_output_name(extension):
     """ generate a unique filename based on the date and time for a given extension.
     """
-    date_time = dt.datetime.now().strftime("%m%d%y_%H-%M")
+    date_time = dt.datetime.now().strftime("%Y%m%d_%H%M%S")
```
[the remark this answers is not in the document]
mimischi: `%m%d%y`, e.g., 122418, is a horrible format. We should use the standard `%Y%m%d`.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,185 @@
...
+    plt.tight_layout()
+    # here I return the figure as well as the legend
+    return ax, lgd
+
+
+def filter_dataframe_for_plotting(df, host_name, module_name, gpu, cpu):
+    # gpu/cpu can be plotted together or separately
+    if gpu and cpu:
+        # if no flags are given by the user or both are set everything is plotted
+        console.info("Plotting GPU and CPU data.")
```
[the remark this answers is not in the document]
mimischi: I think verbosity is great. This reaffirms that we are doing what the user requested.

mdbenchmark/analyze.py
mimischi: We should make up our mind here. We do not use a common style for these messages. In general all commands should be defined via `kwargs`, such that they are shown in bold letters. We could think about also hyphenating them, to aid users with terminals that do not support colors.

mdbenchmark/analyze.py
mimischi: I don't see why using the current approach is bad. If we ever change the labels to something longer, we may end up with two lines anyway.

mdbenchmark/utils.py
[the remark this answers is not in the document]
kain88-de: There is an ISO standard for dates it's `%Y-%m-%d`. That can actually be read by a human.

mimischi:
> How do we best catch the error which is generated by click because nothing is plotted and evaluated is such that the test passes?

I just tried out the current PR. Besides a failing import of `analyze_run`, you do not have a problem with testing the message. It should work with the approach you're using via the `capsys` fixture provided by `pytest`.

For me the test fails because you're not catching the `SystemExit`. Have a look how it is solved for other tests here: https://github.com/bio-phys/MDBenchmark/blob/4e3a8224909500b84b0ec41cfc8c1cd3b05a6b45/mdbenchmark/tests/test_console.py#L65-L73

mdbenchmark/utils.py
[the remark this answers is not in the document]
mimischi: Personally I do not prefer to have hyphens (`-`) inside of filenames, but this is a personal taste. Going for `%Y-%m-%d` or `%y%m%d` should be better than what we have currently.

MSiggel:
@mimischi: the pytest fixes are pushed now. You can have a look if you want. 
I will fix the other comments sometime soon.

mimischi:
@kain88-de Let's rebase this PR once everything is finished. I don't think rebasing it all the time makes much sense.

mimischi:
Closing and re-opening to try and fix the Travis build.

mdbenchmark/analyze.py
mimischi: So we go with something like `console.info('DEPRACATED Please use xyz in the future')`!

mdbenchmark/plot.py
mimischi: `x` would not be correct, because currently `df_sel` is used for selecting the performance data for the y-axis. I think going with something like `selection` or more generally `mask` should be alright. A docstring could help clarify things.

mdbenchmark/plot.py
```diff
@@ -0,0 +1,185 @@
...
+    elif cpu and not gpu:
+        df = df[~df.gpu]
+        console.info("Plotting CPU data only.")
+    elif not cpu and not gpu:
+        console.error("CPU and GPU not set. Nothing to plot. Exiting.")
+
+    for host in host_name:
+        if host in df['host'].tolist():
+            df = df[df['host'].isin(host_name)]
+            console.info('Data for the following hosts will be plotted: {}.', set(host_name))
```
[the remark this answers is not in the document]
mimischi: We should change the whole logic here. I'd propose the following sketch:

1) If no `hostname` is given, we plot all available hosts.
2) If `hostname` is given, we check the names for availability. This results in two cases:
a) One or more provided hostnames can not be found: we use `console.error` to notify the user and exit here.
b) All provided hostnames are available in the CSV/log files: we plot all given hosts.


This would change the code to this:

```python
# Case 1
if not host_name:
    console.info("Plotting all hosts in csv.")

df_filtered_hosts = df[df['hostname'].isin(host_name)]
df_unique_hosts = np.unique(df_filtered_hosts)

# Case 2a
if df_unique_hosts.size != len(host_name):
    console.error("Could not find all provided hosts. Available hosts are: XYZ")

# Case 2b
console.info("Data for the following hosts will be plotted: {}".format(df_unique_hosts))
```

mdbenchmark/plot.py
mimischi: See my comment above.

mdbenchmark/tests/test_plot.py
MSiggel: yes, I am writing the pdf/png file which I also check for at the end of the tests

mdbenchmark/plot.py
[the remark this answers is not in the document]
MSiggel: for now I left the verbosity as is. We could think about adding a feature to our console function which allows for a -v flag or something along these lines..

mdbenchmark/plot.py
MSiggel: I am leaving this as is for now

MSiggel:
I addressed all comments above in some way or left a reply. All tests I wrote are passing and this should be in principle ready for merging after a rebase.
What do you think @mimischi ?

mdbenchmark/analyze.py
```diff
@@ -177,19 +98,14 @@ def analyze(directory, plot, ncores, output_name):
     df.to_csv(output_name)
 
     if plot:
-        df = pd.read_csv(output_name)
+        console.warn('DEPECATED Please use \'mdbenchmark plot\' in the future')
```
mimischi: Minor typo:

```python
console.warn('DEPRECATED Please use \'{}\' in the future.', 'mdbenchmark plot')
```

mdbenchmark/analyze.py
```diff
@@ -177,19 +98,14 @@ def analyze(directory, plot, ncores, output_name):
...
+        fig = Figure()
+        FigureCanvas(fig)
+        ax = fig.add_subplot(111)
 
-        # Fail if we have no values at all. This should be some edge case when
-        # a user fumbles around with the datreant categories
-        if df['gpu'].empty and df[~df['gpu']].empty:
-            console.error('There is no data to plot.')
+        df = pd.read_csv(output_name)
+        ax, lgd = plot_over_group(df, plot_cores=False, ax=ax)
```
mimischi: This does not work. You do not return `lgd` in `plot_over_group`. Instead do it like here:

https://github.com/bio-phys/MDBenchmark/blob/plotting-command/mdbenchmark/plot.py#L176-L178

mdbenchmark/plot.py
```diff
@@ -0,0 +1,190 @@
...
+
+    fig = Figure()
+    FigureCanvas(fig)
+    ax = fig.add_subplot(111)
+    ax = plot_over_group(df, plot_cores, ax=ax)
+
+    lgd = ax.legend(loc='upper center', bbox_to_anchor=(0.5, -0.175))
+    plt.tight_layout()
+
+    if output_name is None:
```
mimischi: Nit picking here: I just noticed that the name of the plot will be different from the CSV file. We should stick with the CSV filename in this case. Otherwise we end up with these seemingly unrelated files: `2018-07-23_140304.csv` and `2018-07-23_140328.png` (note the different seconds).

mimischi:
I squashed everything and rebased onto `develop`. I'll check again tomorrow in the morning and merge it if all is good!

mimischi:
We're pretty much ready. I'd like to go over the strings once more and also make sure that this issue is solved:

https://github.com/bio-phys/MDBenchmark/issues/24#issuecomment-369969273

mimischi:
We're merging as soon as Travis-CI has finished.

Thanks to @MSiggel for the PR!

PR #57, opened 2018-04-08
Refactor common MD engine functions

mimischi:
- Consolidate `write_bench` function from `mdengines.gromacs.write_bench` and `mdengines.namd.write_bench` into `mdengines.utils.write_benchmark`.
- Add unit tests to `mdengines.utils.write_benchmark`. We now also check wheter all variables in the generated `bench.job` were replaced correctly.
- Consolidate `cleanup_before_restart` for both NAMD and GROMACS.
- Consolidate `parse_ns_day`, `parse_ncores` and `analyze_run` for both NAMD and GROMACS.

PR Checklist
------------


mimischi:
Regarding dfce853: I consolidated `parse_ns_day` into `mdengines/utils.py`, but still test the cases separately inside of `tests/mdengines/test_namd.py` and `tests/mdengines/test_gromacs.py`. I somewhat have the feeling we should do that testing inside of `tests/mdengines/test_utils.py`, but we could compromise here to have the tests engine specific.

mimischi:
This should be finished for review.

mimischi:
We now always pass the actual MD engine via `engine` (currently: `mdengines/gromacs.py` or `mdengines/namd.py`) in the code base. Subsequent functions now call `engine.NAME` internally, to access the `NAME` property, like @kain88-de suggested.

We could later think about rewriting parts of the MD engine to use classes, but this should be fine as a first step.

This is ready to be reviewed @MSiggel @kain88-de.

PR #71, opened 2018-05-18
Suppress UserWarning caused by MDAnalysis 0.18

mimischi:
- Filter UserWarning caused by MDAnalysis 0.18

PR Checklist
------------
 - ~[ ] Added changelog fragment in  `./changelog/` ([more information](https://github.com/bio-phys/MDBenchmark/blob/master/DEVELOPER.rst#using-towncrier))?~

### DIFF
--- a/changelog/71.bugfix
+++ b/changelog/71.bugfix
@@ -1,0 +1,1 @@
+Suppress UserWarning caused by ``MDAnalysis==0.18``.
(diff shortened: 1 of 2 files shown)


mimischi:
Actually now that `h5py==2.8` is released, their `FutureWarning` is gone. I think we should leave in that old filter, because users may not update to the newest `h5py` version and we do not want to force a minimal version in the package.

Ping @kain88-de

PR #72, opened 2018-06-18
Error handling

kain88-de:
Fixes #

This PR adds a `console.redirect` context manager. It can be used in click command functions to turn warnings and exceptions into the appropriate click error messages. This PR can help testing because getting error messages from `click.error` or `click.warn` from library functions (not decorated with `@click...`) can use the normal python idioms of exceptions and warnings that are easy to test for.

I have only tested this now by hand. I would like some opinion before I go deeper into this if this change is appreciated. 

The current approach to the redirect is very 'holistic'

```python
@cli.command()
def command(...):
    with console.redirect():
        ...
```

But it can/should be used more  targeted in the functions

```python
@cli.command()
def command(...):
    ...
    with console.redirect():
        library_func_with_exceptions
    ...
```

PR Checklist
------------


kain88-de:
So one problem with the current 'holistic' approach is that warnings are all shown at the end. I also have no good solution for the `console.info` case.

PR #79, opened 2018-07-26
Update Travis-CI runs (and reformat code with black)

mimischi:
- Use the caching functionality provided by Travis-CI to cache downloaded Python packages. This will still re-install the packages every time, but does not need to re-download them.
- Run Travis-CI jobs in job steps. We stopped using [`ci-helpers`](https://github.com/astropy/ci-helpers) for most of the jobs, because one run can take between 4-8 minutes. This way we can get feedback from Travis very quickly and do not have to wait for the extremely slow jobs to finish. Also code linting is done outside of the unit test jobs -- so now unit tests can only fail if the actual tests fail. The new approach only needs 2 minutes.
We now do the following:
1) Run unit tests on Linux with Python 3.6, 3.5 and 2.7. If these fail, stop.
2) Run `black --check`, `flake8` and `isort --check-only` to check for formatting and linting errors. Also lint our `README.rst`. If these fail, stop.
3) Install `ci-helpers` inside a macOS environment and run unit tests in Python 2.7.

- Add `black` to `.travis.yml`, to check for formatting issues. Builds will fail, if `black` would reformat any files inside of `mdbenchmark/`.
**Note for developers:** Go and install `black` via `pip install --user black` onto your system and make sure to run it, before submitting code. `black` integrates into a variety of editors, see [`black` README](https://github.com/ambv/black/blob/master/README.md#editor-integration). If possible, configure it to run after saving your python file.
- Set `flake8` to ignore some annoying things, e.g., E501: line too long. `black` will take care of reformatting and should do its job accordingly. We should not be bothered to correct such things anymore. Also removed the `max-line-length` option from `setup.cfg`.
- Update `Makefile` and add some new management commands (like `make reformat` to run `black` and `make flake8` to run the `flake8` check).

PR Checklist
------------
 - ~[ ] Added changelog fragment in  `./changelog/` ([more information](https://github.com/bio-phys/MDBenchmark/blob/master/DEVELOPER.rst#using-towncrier))?~


mimischi:
Integration with [`isort`](https://github.com/timothycrosley/isort) would still be nice. I really like having the imported packages sorted in [a predictable way](https://github.com/timothycrosley/isort#how-does-isort-work).

mimischi:
Pinging @kain88-de for feedback. We're now enforcing a _very_ opinionated formatting of the code, but this will lead to consistent formatting and basically no overhead for the developer to decide how to format things to make pep8 happy.

There is only one thing that Travis-CI does not seem to like, but I don't know what's wrong with `.travis.yml`. It's the `codecov` line at the bottom, but the syntax looks alright: https://travis-ci.org/bio-phys/MDBenchmark/jobs/408393820#L8128

PR #100, opened 2018-10-11
Add documentation to develop

mimischi:
This is convoluted, but we also want to have the documentation in `develop`, before merging it into `master`.

For all discussions see PR #88.

PR #88, opened 2018-09-24
Add documentation

mimischi:
This PR adds a documentation to the project. Replaces #84.

The current version is hosted on readthedocs.io: https://mdbenchmark.readthedocs.io/en/latest/

PR #101, opened 2018-10-11
Add documentation to develop

mimischi:
This is convoluted, but we also want to have the documentation in `develop`, before merging it into `master`.

For all discussions see PR #88. Replaces #100.


mimischi:
Something went wrong with #100. Testing changes here.

mimischi:
@kain88-de Any pointers on why the OS X job is failing?

https://api.travis-ci.org/v3/job/440162835/log.txt

.travis.yml
```diff
@@ -60,15 +61,18 @@ jobs:
...
     - stage: "macOS - Unit tests"
       language: generic
       name: "Python 2.7"
       env: PYTHON_VERSION=2.7
-      install:
-        - git clone git://github.com/astropy/ci-helpers.git
-        - source ci-helpers/travis/setup_conda.sh
-        - pip install -e .
-      script: $MAIN_CMD $SETUP_CMD
+      install: pip install . mdbenchmark -v
```
kain88-de: This build runs now on ubuntu and *not* osx. Could this be part of the problem? You sed the PYTHON_VERSION environment variable but the global used version is 3.6 set at the top of the file.

mimischi: Wow, that makes sense. Solving month-old merge conflicts is a pain. Thanks for the heads up!

PR #114, opened 2018-10-21
initial implementation of lammps support

MSiggel:
Initial support for lammps simulations. 
lammps module added, parsing added and test files added. 

PR Checklist
------------

### DIFF
--- a/mdbenchmark/generate.py
+++ b/mdbenchmark/generate.py
@@ -31,7 +31,11 @@ NAMD_WARNING = (
     "Parameter paths must be absolute. Only crude file checks are performed! "
     "If you use the {} option make sure you use the GPU compatible NAMD module!"
 )
-
+LAMMPS_WARNING = (
+    "LAMMPS support is experimental."
+    "All input files must be in the same directory and have the same base name!"
+    "We hope you know what you're doing..."
+)
 
 def validate_name(ctx, param, name=None):
     """Validate that we are given a name argument."""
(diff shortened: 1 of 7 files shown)


MSiggel:
I added most of the stuff to use the code with LAMMPS in a crude manner. Same as in namd an issue here is that multiple files are required to run a simulation. Similiar issue will be the case when using AMBER. In light of this we should also discuss in this PR how we handle files which have to be provided i.e. whether we just assume that the user has to provide all files with the same name or not.
This would decrease reliability for those engines because the user might make mistakes.

mdbenchmark/mdengines/lammps.py
```diff
@@ -0,0 +1,83 @@
...
+        # Continue if we do not need to do anything with the current line
+        if (
+            ("parameters" not in line)
+            and ("coordinates" not in line)
+            and ("structure" not in line)
+        ):
+            continue
+
+        path = line.split()[1]
+        if "$" in path:
```
mimischi: This is all NAMD specific.

mdbenchmark/generate.py
```diff
@@ -31,7 +31,11 @@
     "Parameter paths must be absolute. Only crude file checks are performed! "
     "If you use the {} option make sure you use the GPU compatible NAMD module!"
 )
-
+LAMMPS_WARNING = (
+    "LAMMPS support is experimental."
+    "All input files must be in the same directory and have the same base name!"
+    "We hope you know what you're doing..."
```
mimischi:
```suggestion

```

mdbenchmark/mdengines/lammps.py
```diff
@@ -0,0 +1,83 @@
+# -*- Mode: python; tab-width: 4; indent-tabs-mode:nil; coding:utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4 fileencoding=utf-8
+#
+# MDBenchmark
+# Copyright (c) 2017 Max Linke & Michael Gecht and contributors
```
mimischi:
```suggestion
# Copyright (c) 2017-2018 The MDBenchmark development team and contributors
```

MSiggel:
I added the ```analyze-files-lammps``` folder. There is a sample log file in folder ```/1```.
I'll have to rerun this simulation the cluster.

PR #118, opened 2018-10-23
Update documentation for v2.0

### LINKED ISSUE #103: Update documentation for 2.0
We just merged the documentation into `develop` (PR #101). Before releasing version 2.0, we should update the documentation to reflect the new functionality. This should actually be only minor changes, because the biggest part was to write the initial documentation in the first place.

MSiggel:
Fixes #103 

Updates the documentation for version 2.0 
A new section is added for the plotting command.

PR Checklist
------------


mimischi:
@MSiggel Could you please have a look at this?

docs/analyze.rst
```diff
@@ -49,23 +50,22 @@ single benchmark::
 Plotting of benchmark results
 -----------------------------
 
+  .. warning::
+
+   The function |mdbenchmark.analyze.plot|_ was deprecated with version 2.0. You should migrate to the newer ``mdbenchmark plot``, as it provides more functionality and the former version will be removed in the future.
+
 MDBenchmark provides a quick and simple way to plot the results of the
```
MSiggel: But this still has the problem that multiple modules and gpu/cpu will not pbe handled correctly. We should make clear that the functionality is very limited and in a way "buggy"

docs/generate.rst
```diff
@@ -121,6 +121,14 @@ HPCs. To change the run time per benchmark, simply use the ``--time`` option::
...
+---------------------
+
+If you want your benchmark jobs to have specific names, use the ``--job-name`` option::
+
+  mdbenchmark generate --job-name my_benchmark
+
 .. _modules: https://linux.die.net/man/1/module
 .. _draco: https://www.mpcdf.mpg.de/services/computing/draco
 .. _hydra: https://www.mpcdf.mpg.de/services/computing/hydra
+.. _datreant: https://www.datreant.org/
```
MSiggel: I added a short section on datreant which seems to be gone now. Should we leave this in as reference even though it is not mentioned in the text?

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Output options and file formats
+-------------------------------
+
+You can set the output name for the generate plot using the ``--ouput-name`` option. If no name is given, the current date and time will be used as a file name. To generate a plot with the filename ``my_benchmark_plot``, simply run::
+
+   mdbenchmark plot --output-name my_benchmark_plot
+
+You can also specify a filetype with the ``--output-format`` option.
+Supports file formats are ``png``, ``pdf``, ``svg`` and ``ps``. To set the file format to PNG, run::
```
MSiggel: use pdf as example because png is the default.

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+   mdbenchmark plot --csv data1.csv --csv data2.csv
+
+This will plot all data from the benchmark results found in the given files. These can be filtered using the options detailed below. All filters are applied to data in all CSV files collectively.
+
+Output options and file formats
+-------------------------------
+
+You can set the output name for the generate plot using the ``--ouput-name`` option. If no name is given, the current date and time will be used as a file name. To generate a plot with the filename ``my_benchmark_plot``, simply run::
+
+   mdbenchmark plot --output-name my_benchmark_plot
```
MSiggel: by default this will generate a my_benchmark_plot.png file.

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+
+You can also specify a filetype with the ``--output-format`` option.
+Supports file formats are ``png``, ``pdf``, ``svg`` and ``ps``. To set the file format to PNG, run::
+
+   mdbenchmark plot --output-format png
+
+Filter what to plot
+-------------------
+
+To filter your data from the given CSV file(s) you can use the following commands. These can be mixed together as you like.
```
MSiggel: These can be combined as you like.

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+   mdbenchmark plot --no-gpu
+
+To create a plot without CPU benchmarks run::
+
+   mdbenchmark plot --no-cpu
+
+Filter by MD engine
+~~~~~~~~~~~~~~~~~~~
+
+If you have run benchmarks for different MD engines, but want to only plot a subset of these, you can do so with the ``--module`` option. For example, if you wanted to plot all benchmarks for the two modules ``gromacs/2018.2`` and ``gromacs/2016.4``, you would run::
```
MSiggel: if you want [...] run (omit the would for consistency)

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+The plot resolution can be changed with the ``--dpi`` option. The default ist ``300``::
+
+  mdbenchmark plot --dpi
+
+Customizing ticks on the x-axis
+-------------------------------
+
+It is possible to change the frequence of ticks on the x-axis. To do this, call the ``--xtick-step`` option::
+
+  mdbenchmark plot --xtick-step
```
MSiggel: mdbenchmark plot --xtick-step  5?
 Maybe add integer value example which would give reasonable results.

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Removing the watemark
```
MSiggel: Removing the watermark

docs/plot.rst
```diff
@@ -0,0 +1,111 @@
...
+---------------------
+
+By default a small watermark will be placed in the top left corner of every plot. To disable this, use the ``--no-watermark`` option::
+
+  mdbenchmark plot --no-watermark
```
MSiggel: we should add a note that it would be nice if users left this or cited the zenodo DOI to give us credit when they use the tool.

docs/analyze.rst
mimischi: No, for this we have the old documentation that describes exactly the use case. Users should read the documentation for the version they are using.

docs/generate.rst
mimischi: I don't think it is necessary to say anything about `datreant` at the current point.

mimischi:
@MSiggel This should be finished.

MSiggel:
@mmischi looks good!

PR #119, opened 2018-10-23
Update analyze CSV output settings

### LINKED ISSUE #95: Saner defaults for CSV output generation and plotting
With the recent plotting revamp (PR #52), we generate a new CSV file everytime a user runs `mdbenchmark analyze`. The file name is generated from the current date and time, resulting in something like `2018-05-24_185930.csv`. If a user is constantly spamming `mdbenchmark analyze`, they will get a new CSV file indefinitely.
Additionally, if the user wants to plot a CSV file they provide the name as an argument to `mdbenchmark plot --csv`.

With this I would like to propose:

1) Do not generate a CSV file automatically, unless asked to. `mdbenchmark analyze` can then be used to just check on the benchmarks quickly, without generating a dozen CSV files.
- Add a new option `mdbenchmark analyze --csv <optional_name>`. If given, we will generate a CSV file and tell the user.
2)  If `mdbenchmark plot` is called without the `--csv` option, we scan the current working directory for `.csv` files and let the user choose which to plot.
3) If given a `--csv` option, plot the file like we do currently.

I am still not 100% convinced by our current naming scheme for CSV files, but given that we provide the user with an option to change the default, we should actually be fine for now.

MSiggel:
Fixes #95 

- saner csv output names
- default behavior of analyze changed such that by default no csv is saved.
- when invoking ```--save-csv``` a file is saved

PR Checklist
------------

### DIFF
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -60,13 +60,12 @@ plt.switch_backend("agg")
     show_default=True,
 )
 @click.option(
-    "-o",
-    "--output-name",
+    "-s",
+    "--save-csv",
     default=None,
     help="Filename for the CSV file containing benchmark results.",
-    type=str,
 )
-def analyze(directory, plot, ncores, output_name):
+def analyze(directory, plot, ncores, save_csv):
     """Analyze benchmarks and print the performance results.
 
     Benchmarks are searched recursively starting from the directory specified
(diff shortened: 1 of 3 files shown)


mdbenchmark/analyze.py
```diff
@@ -94,12 +93,9 @@ def analyze(directory, plot, ncores, output_name):
     # move this to the bundle function!
     PrintDataFrame(df)
 
-    # here we determine which output name to use.
-    if output_name is None:
-        output_name = generate_output_name("csv")
-    if ".csv" not in output_name:
-        output_name = "{}.csv".format(output_name)
-    df.to_csv(output_name)
+    if save_csv is not None and ".csv" not in save_csv:
```
mimischi: Just realized this will actually allow files with the name `my_output.csv.something`. Instead we should do:

```suggestion
    if save_csv is not None and not save_csv.endswith(".csv"):
```

mdbenchmark/analyze.py
```diff
@@ -59,13 +59,12 @@
     show_default=True,
 )
 @click.option(
-    "-o",
-    "--output-name",
+    "-s",
+    "--save-csv",
     default=None,
-    help="Filename for the CSV file containing benchmark results.",
-    type=str,
```
mimischi: What happens if we call `mdbenchmark analyze --save-csv 1`? Will this be handles as an `int`, leading to a crash in the `if`-clause on line 96? Or does `click` handle this correctly, converting `1` to a `str`?

MSiggel: click handles this correctly.

PR #123, opened 2018-10-23
updated draco template

### LINKED ISSUE #116: Update draco template
The job templates for the MPCDF cluster `draco` were changed and we should update our default one. It is not possible to run jobs on 1 _or_ 2 GPUs on `draco`, but I would suggest to keep the current way of requesting all available GPUs, as we currently do not have the functionality to support the other way.

https://www.mpcdf.mpg.de/services/computing/draco/sample-batch-script

MSiggel:
Fixes #116

draco template updated to the new specifications by MPCDF

PR Checklist
------------

### DIFF
--- a/mdbenchmark/templates/cobra
+++ b/mdbenchmark/templates/cobra
@@ -31,6 +31,7 @@
 #SBATCH --time={{ formatted_time }}
 
 module purge
+module load intel
 module load impi
 module load cuda
 module load {{ module }}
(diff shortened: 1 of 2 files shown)


mimischi:
After doing `module purge`, we now need to also re-load the `intel` module via `module load intel`:

```bash
module purge
module load intel
module load impi
module load cuda
module load {{ module }}
```

mimischi:
Oh, while you're at it: the same change needs to be done to `cobra`. Actually also to `hydra`, but I don't know which modules we need to load there right now.

PR #126, opened 2018-10-23
Backwards compatible --plot--cores

### LINKED ISSUE #113: ncores flag not used used in analyze.py
As far as I can tell the ncores flag is not used in the code and we are always guessing the number of cores. I would suggest we either remove this flag or fix its functionality.

mimischi:
Fixes #113

- Pass value of `--plot-cores` to `mdbenchmark analyze --plot`.

PR Checklist
------------

### DIFF
--- a/mdbenchmark/analyze.py
+++ b/mdbenchmark/analyze.py
@@ -110,7 +110,7 @@ def analyze(directory, plot, ncores, save_csv):
         ax = fig.add_subplot(111)
 
         df = pd.read_csv(save_csv)
-        ax = plot_over_group(df, plot_cores=ncores, ax=ax)
+        ax = plot_over_group(df, plot_cores=ncores, fit=True, ax=ax)
         lgd = ax.legend(loc="upper center", bbox_to_anchor=(0.5, -0.175))
 
         fig.tight_layout()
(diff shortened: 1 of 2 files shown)


mimischi:
This is really not optimal, but I guess we should just go with the current functionality. We do not allow the users to specify any custom number of cores any more and always try to parse from the log file.

mdbenchmark/analyze.py
```diff
@@ -110,7 +110,11 @@ def analyze(directory, plot, ncores, save_csv):
         ax = fig.add_subplot(111)
 
         df = pd.read_csv(save_csv)
-        ax = plot_over_group(df, plot_cores=False, ax=ax)
+        if ncores:
+            console.warn(
+                "Ignoring any values you specified and parsing number of cores from log files."
```
MSiggel: shouldn't this be the other way around?
```if not ncores:```
If the user gives a value for ncores the value overrides the automated parsing.

mimischi: That is what it did before. The current implementation (re-write for `mdbenchmark plot`), only accepts `--plot-cores` as an option and then plots the number of cores, instead of nodes.

This does lead to the problem: if we cannot extract number of cores values from certain log files, but the performance is there, then we will not plot these values.

PR #128, opened 2018-10-24
Add python environment check

### LINKED ISSUE #127: Upgrading to new MDBenchmark version breaks environment
With our upgrade to `datreant>1.0`, we need to tell our users to uninstall `datreant.core` and `datreant.data` from their Python/Conda environments. Otherwise execution of `mdbenchmark` will not work.

I wrote a small workaround, where we use `console.error` to warn the user, that their current environment is incompatible with MDBenchmark v2.0 and they should delete those packages.

This is really ugly, but I guess there is no other way around this.. @kain88-de?

mimischi:
Fixes #127

- Added a function that tries to import `datreant.core`, `datreant.data` and `mdsynthesis` and throws an error at the user, asking to uninstall these python packages.

PR Checklist
------------


mimischi:
Why is this failing on Travis but working locally..

mimischi:
So this works, but I still like to add a test for `mds_to_dtr.ensure_correct_environment()`.

PR #136, opened 2018-10-29
Merge master into develop, into master

mimischi:
Let's try if this works.

We performed a merge of `master` into `develop` and then merged `develop` back into `master via `git merge --no-ff develop`.


mimischi:
@kain88-de Is this the correct way to do this? We got into this situation, because we merged some stuff directly into `master`, without going through `develop` first.

PR #150, opened 2019-07-22
fix: update AliasedGroup CLI test

### LINKED ISSUE #115: Update Click to 7.0
Click v7.0 breaks some of our tests. We should fix those tests, so we can upgrade to `click>=7` at some point.

mimischi:
Fixes #115

- Updated test for `AliasedGroup`.

PR Checklist
------------

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 .eggs/
 build/
 dist/
+pip-wheel-metadata/
 
 # VSCode
 .vscode/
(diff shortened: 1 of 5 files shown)


mimischi:
So this finally resolves #115 and we therefore should be able to also get https://github.com/conda-forge/mdbenchmark-feedstock/pull/12 out of the way.

Unfortunately, we are stuck with `xdg<=1.0.4` as long as we support Python 2. I would propose to drop it soon™️.

PR #153, opened 2019-08-01
[WIP] Refactor CLI

### LINKED ISSUE #152: CLI is slow
**Expected behavior**

Any `mdbenchmark` command should print output reasonably fast.

**Actual behavior**

It sometimes takes ~5 seconds for any output to appear. This is most likely due to this line:

https://github.com/bio-phys/MDBenchmark/blob/49b8f44585fd045ca3779694bd8ecab99d80a4ad/mdbenchmark/__init__.py#L20

Here, we import all CLI subcommands, which in their turn import different libraries like `pandas`, `matplotlib.pyplot`, etcpp. When calling `mdbenchmark generate` or `mdbenchmark submit` we most certainly are not interested in importing `matplotlib`.

**Your version of MDBenchmark:**

2.0.0

mimischi:
Fixes #152. I took inspiration in how `pipenv` solves this: https://github.com/pypa/pipenv/tree/master/pipenv/cli

- Updated how we initialize the CLI.

PR Checklist
------------
 - [ ] Migrate tests from old to new files. This should fix the coverage discrepancy.


mimischi:
I performed some benchmarks using Pythons 3.7 [`importtime` profiler](https://docs.python.org/3/using/cmdline.html#id5) with `python -X importtime mdbenchmark generate 2> import.log.log`

Results can be viewed on [this neat Netlify instance](https://mdbenchmark-importtime.netlify.com/), visualized with [`tuna`](https://github.com/nschloe/tuna).

# Results

I am aware, that I am _not_ importing 100% of the packages that are currently imported in `develop`. Nevertheless, I am importing all external libraries, which lead to the long import times.

- `mdbenchmark` is 5.5x faster (0.168s instead of 0.923s).
- `mdbenchmark analyze` is 1.07x slower (0.923s instead of 0.858s).
- `mdbenchmark generate` is 1.44x faster (0.562s instead of 0.807s).

Not sure what is happening with `mdbenchmark analyze`, but given that we do _not_ import `matplotlib` when calling `mdbenchmark generate` is a great speedup.

mimischi:
We could further refactor the code and make `mdbenchmark` useable through a Python API. However, I think to make this concise, it will require some work as well as a new class-based structure?

mimischi:
Re-did the analysis of the `cli` branch using the script from [this gist](https://gist.github.com/mimischi/c3ee9de7692aa570b5ac816f3fe82391).

[Results on Netlify](https://mdbenchmark-importtime.netlify.com/)

# TL;DR
Most commands are **way** faster, `mdbenchmark submit` and `mdbenchmark plot` are somewhat slower (`n=1`).

# Results

Command | Branch: `develop` (s) | Branch: `cli` (s) | Fold change
------------ | ------------- |  ------------- |  -------------
`mdbenchmark` | 2.741 | 0.159 | 17.24
`mdbenchmark generate` | 1.248 | 0.143 | 8.73
`mdbenchmark submit` | 0.792 | 0.851 | 0.93
`mdbenchmark analyze` | 2.188 | 0.771 | 2.84
`mdbenchmark plot` | 0.812 | 0.871 | 0.93

## Limitations
Some commands, like `mdbenchmark generate`, exit with an error and we therefore do _not_ consider the whole function. Nonetheless, this is true for both `develop` and this `cli` branch, so the comparison should be rather fair. And although the command exits early, the startup time has decreased by almost 9 fold.

mimischi:
We are having problems with three tests in `test_submit.py`, because of our monkeypatching. I marked them to be skipped for now, because I want this refactor in now. We might need to refactor some code to get rid of the monkeypatching or re-think how we test the CLI.

PR #165, opened 2020-04-29
Add ranks and threads to templates

### LINKED ISSUE #44: benchmarking procedure improvements sugestions
**Expected behavior**
I/O, load balancing, CPU/GPU throttling and other performance influencers need to be accounted for by using timer resetting.
See Benchmarking procedure section of [10.1002/jcc.24030](http://onlinelibrary.wiley.com/doi/10.1002/jcc.24030/full)

**Actual behavior**
Timer seems to never be reset.

**Your version of MDBenchmark:**
(run `mdbenchmark --version` on your machine)

### LINKED ISSUE #60: Run with different OpenMP/MPI configurations
Performance can be hugely influenced by the composition of MPI processes to OpenMP threads in gromacs and other engines. For this, it would be helpful to allow setting the ratio of OpenMP threads to MPI processes. Currently, we limit the number of MPI processes to the number of cores. So an option should select the ratio of hardware cores to MPI processes. Take the following mock command.

`mdbenchmark generate .... --core_to_mpi_ratio= 1 2 4 8`

With this, we set up 4 simulations. The ratio is defined as `nprocesses = ncores / ratio`. So for a machine with 24 cores and a ratio of 2, we would use 12 MPI processes. For an equal division, a ratio value should be a power of 2. I'm open for better names to this flag.

Also good would be an option to set up a run without MPI only using OpenMP on a single node.

mimischi:
## Rationale

We want to allow users to change the number of MPI ranks and OpenMP threads in their benchmarks, because these settings affect the simulation performance.

### Validation of `--ranks`
The validity of the `--ranks` option is determined on the current machine. If the number of physical cores is different on the login and work nodes and the user specifies a number of ranks (`--ranks`) that would be valid on the work nodes we throw an error.

~We can probably solve this by adding another `--threads` option, in which we ignore all our own calculations and use user specified values? I would first go with this leaner implementation and see whether anyone complains.~

I added the `--physical-cores` and `--logical-cores` options to mitigate this issue. I don't like this solution, but there is probably no way around this.

### Backwards compatibility
We changed the folder structure from `cobra_gromacs/2019.6/5` to `cobra_gromacs/2019.6/n005_r40_t02_wht`. This is more explicit, but not backwards compatible in case users have written some bash scripts that assume a certain folder naming.

I am proposing to release this feature with MDBenchmark v3.0, where we also deprecate Python 3. We might also want to refactor other parts of the code, because there is some inconsistencies that could be improved.

## Changelog

- Add `--ranks` option, with which the user can specify the number of MPI ranks to use. The number of OpenMP threads will be calculated accordingly. Resolves #60.
- Add `--hyperthreading` option to toggle usage of hyperthreading.
- Add `--physical-cores` and `--logical-cores` to specify the number of cores on the work nodes. If `--logical-cores` is not specified, will assume hyperthreading is possible (`logical-cores = 2 * physical-cores`). If `--physical-cores` is not specified, will retrieve information from the current machine.
- Update MPCDF cluster `cobra` template to include (not so) recent changes to the queuing system.
- Add ranks/threads/hyperthreading variables to `cobra` and `draco` templates.
- Add `-resethway` option to `mdrun` command. Resolves #44.

PR Checklist
------------
 - [ ] Update documentation
 - [ ] Update examples in `README.md`


mimischi:
Argh! The login nodes on `cobra` have 48 logical cores, making this implementation useless.

mimischi:
Merging this into the `version-3` branch.

PR #170, opened 2020-10-09
Release 3.0.1

mimischi:
This is the release of version 3.0.1. It includes all functionality as described in [doi:10.1063/5.0019045](https://doi.org/10.1063/5.0019045)


mimischi:
Whoops, force pushed instead of merging. My bad.

